PR #4, opened 2020-10-16
Use React of the parent project instead of our own

cprecioso:
`react-notion-x` should not use its own React, but its parent's, as using two different React copies might cause problems. While usually package managers deduplicate dependencies, in some cases they might choose not to (and the semantics of `node`'s package system don't force them to deduplicate either). Declaring `react` and `react-dom` as peer dependencies will make clear that they have to use the parent's copy, and declaring it in the `devDependencies` will make sure that they are available for local development and testing.


transitive-bullshit:
Thanks @cprecioso!

transitive-bullshit:
Just published as `v2.5.1`.

PR #12, opened 2021-01-03
Update styles.css changed svg.notion-page-icon to object-fit: fill

RodriguesCosta:
change object-fit in svg.notion-page-icon

using object-fit cover
image

using object-fit fill
image


transitive-bullshit:
Looks good -- thanks!

PR #16, opened 2021-01-25
Detect if an anchor link is external and link set rel and target accordingly

dstollie:
With this PR we we check whether an anchor links to an internal or external page. On our notion page we use a links to link a internal pages.

Example https://service-zonneplan.vercel.app/b5830d7f3bd04a7788b819461a998b69 with the links "Ga naar". We use a link instead of pageLink because we want to have the link a different title than the page itself.

The check is pretty simple; if the anchor contains http or https, we decided that is is in an external page.

### DIFF
--- a/packages/react-notion-x/src/components/text.tsx
+++ b/packages/react-notion-x/src/components/text.tsx
@@ -134,7 +134,7 @@ export const Text: React.FC<{
               return <Equation math={decorator[1]} />
 
             case 'a':
-              const isExternalLink = !!linkProtocol;
+              const isExternalLink = !!linkProtocol
 
               return (
                 <components.link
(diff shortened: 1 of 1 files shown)


transitive-bullshit:
Hey @dstollie, I think I solved this a different way here https://github.com/NotionX/react-notion-x/blob/master/packages/react-notion-x/src/components/text.tsx#L141

I think it covers your use case, but please let me know if it doesn't.

normdoow:
> Hey @dstollie, I think I solved this a different way here https://github.com/NotionX/react-notion-x/blob/master/packages/react-notion-x/src/components/text.tsx#L141
> 
> I think it covers your use case, but please let me know if it doesn't.

@transitive-bullshit how does this work exactly? I'm still seeing anchor links open up in a separate page.

PR #25, opened 2021-02-04
fix no value at content block id

nabettu:
Some notion pages have this value[1] empty. At that time, id is undefined and an error will occur in the subsequent processing, so I tried not to add it if it is empty.


transitive-bullshit:
Thanks! Would you mind including a test page that reproduces the issue?

nabettu:
Oh sorry. @transitive-bullshit 
this is error page.

https://www.notion.so/nabettu/Error-Page-9b95bcb5862a495993403d5f8ce99f2b

plz check this by "getPageContentBlockIds".

transitive-bullshit (approved):

transitive-bullshit:
Looks good -- thanks @nabettu 🙏

transitive-bullshit:
Published as `4.2.11`

PR #31, opened 2021-02-18
Typer Error cannot read property 'title' of undefined

normdoow:
Fix empty Notion toggles and possible empty image titles

Public notion page example: https://www.notion.so/Test-Empty-130868ef58d949f4add8cf1da983fc16
Id: 130868ef58d949f4add8cf1da983fc16

With out this change the above notion page would get the error:

Typer Error cannot read property 'title' of undefined

image


transitive-bullshit:
Awesome — thanks, Noah 🙏

PR #33, opened 2021-02-23
fixed block.properties being undefined in collection-card

normdoow:
This issue is related to the last PR in a case when properties is null. Just skipping that property when there is no data.

Happens on this page: https://www.notion.so/ad83934d2fe94e8dba592c8547d6a026?v=7a75440935c44dc88e4dcfcf2b87f28f

image

### DIFF
--- a/packages/react-notion-x/src/components/collection-card.tsx
+++ b/packages/react-notion-x/src/components/collection-card.tsx
@@ -108,7 +108,7 @@ export const CollectionCard: React.FC<CollectionCardProps> = ({
       }
     }
   }
-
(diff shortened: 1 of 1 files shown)


transitive-bullshit:
Hey @normdoow, just make sure the prettier tests pass and I'll merge.

Thanks!

normdoow:
@transitive-bullshit Hadn't used prettier before. But got it installed and working. Should be good to go now!

transitive-bullshit:
No worries :) Published as `4.2.23`

PR #35, opened 2021-02-25
Fixed image height when the image is in a column

normdoow:
Issue here: https://github.com/NotionX/react-notion-x/issues/34#issuecomment-785559577

Screen Shot 2021-02-24 at 8 15 08 PM

Other Images still look good 
Screen Shot 2021-02-24 at 8 16 42 PM


transitive-bullshit:
Thanks, Noah!

transitive-bullshit:
Published as `v4.2.24`

PR #37, opened 2021-03-01
fixed width for columns in browsers such as safari

normdoow:
Fixed columns for safari and other browsers.

Notion page id for example: 067dd719a912471ea9a3ac10710e7fdf

After some playing around I found that px was causing safari to fail calculating the width where as chrome was smarter than that. 😂


transitive-bullshit:
Amazing haha; Thanks, Noah! 🙏

transitive-bullshit:
Published as `v4.3.1`.

PR #45, opened 2021-03-19
Added links to pages in table rows

normdoow:
Currently react-notion-x doesn't have pages for each row in a table. I added linking to a page in a table collection view.
Here is a little video showing the new functionality: https://www.loom.com/share/15cd7ec66aaf466eb0e6699812e1cd85

This more closely matches how Notion behaves with tables.

Tested on e5b45193b4c1436fbde24cad8e9ababd has a table in there among other things.

### DIFF
--- a/packages/react-notion-x/src/components/property.tsx
+++ b/packages/react-notion-x/src/components/property.tsx
@@ -24,7 +24,7 @@ export const Property: React.FC<{
   collection?: types.Collection
   inline?: boolean
 }> = ({ schema, data, block, collection, inline = false }) => {
-  const { components, mapImageUrl } = useNotionContext()
+  const { components, mapImageUrl, mapPageUrl } = useNotionContext()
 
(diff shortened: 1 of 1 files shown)


transitive-bullshit:
Ahhh this used to work but must've gotten broken at some point.

Much appreciated, @normdoow 🙏 😃

transitive-bullshit:
Published as `4.3.5`.

I had hesitated because Property components are displayed from multiple places, and I don't think we always want to make the `title` linkable, but I don't have time to look into counter-examples and this is a good change, so I wanted to get it in & we can always fix edge cases later on.

Thanks again!

normdoow:
@transitive-bullshit, Yeah I was thinking the same thing. I was looking at the other usages and it seemed to me like it would be ok if it was linkable. Only uses the link if it is a page so couldn't think of a case we wouldn't want that.

PR #46, opened 2021-03-20
Added support for Wistia embed videos

normdoow:
Id of page to test: cfdc513910024e91b872648eda848986

Pretty straight forward change!


transitive-bullshit:
Danke; published as `4.3.6`

PR #49, opened 2021-03-30
fix get text to exclude marks for TeX and mention

Niinjoy:
A text content in Notion may include '⁍' for TeX equations and '‣' for mentions. So I changed the code a little bit to exclude this two marks. The `getTextContent` function is used in may places like toc, formula, title of collection, etc. This commit will make those elements display better.

Sample Notin Page ID: e2e6389a160e406d88e312f12ede26e4

By the way, I really appreciate your remarkable work! Just started to use react-notion-x, I hope I can contribute a little bit to this great project.


transitive-bullshit:
Thanks @Niinjoy; this looks great. I just need to find some time to properly test it before merging.

transitive-bullshit:
Published as `v4.6.0`.

PR #57, opened 2021-04-23
Added alt text to images from the caption

normdoow:
Alt text from a caption just wasn't added to the image in lazy-image. Added it in this PR.
The preview image was already good to go.

Can see it tested out in the test suite page id: 3492bd6dbaf44fe7a5cac62c5d402f06


transitive-bullshit:
Looks good 😄

transitive-bullshit:
Published as `v4.5.1`.

PR #61, opened 2021-05-01
Fix text being empty when there is a comment on the block.

normdoow:
Here is a page to test where I saw the issue: 4d967aaf8f204b30a0f50c74f2801987

You can see down about half way the FAQ title and first title of the toggle aren't showing. I found this was because the page had comments there.

The change I made was make the "m" decoration (comments) still return the element instead of skipping it.


normdoow:
@transitive-bullshit, I just wanted to make sure I wasn't missing something here. I did see that some other examples with comments were working before this change. Not sure what the difference is between those.

transitive-bullshit:
LGTM 👍

transitive-bullshit:
Published as `v4.6.0`.

PR #71, opened 2021-05-21
fix property error if no language is set in code block

AlanDecode:
If no language is set in notion code block, this line will yield an error:

```
Cannot read property '0' of undefined
```

See https://github.com/craigary/nobelium/issues/130


transitive-bullshit:
Thanks! Published as `4.6.2`

Nobelium looks awesome btw 💯

PR #72, opened 2021-05-22
Comment out debug logs of collection views

xissy:
Not sure it is intended, collection view components print out some debugging info on browsers or server logs. Hope it is able to control its verbose level in the future. Gently suggesting commenting those out in production.

image
image


transitive-bullshit:
Thanks, @xissy 🙏

transitive-bullshit:
Published as `v4.6.1`

PR #74, opened 2021-05-27
fixed groupBlockContent trying to access index -1

normdoow:
Interesting issue in grouping block content where a block is missing so the value is undefined. Somehow its getting past the null check so then trys to access the array at index -1

happening on this notion page: b1e94e2308b24476a323fcaa6fa1a906


transitive-bullshit:
Published as `v4.6.3` — thanks! 🙏

PR #77, opened 2021-05-31
Update collection property_visibility type.

nabettu:
collection visibility has "hide_if_empty".

スクリーンショット 2021-05-31 9 32 12

### Test Page
https://www.notion.so/nabettu/Page-1-9246e732a60a43c9abc367414b557dca

### DIFF
--- a/packages/notion-types/src/collection.ts
+++ b/packages/notion-types/src/collection.ts
@@ -46,7 +46,7 @@ export interface Collection {
     }>
     property_visibility?: Array<{
       property: PropertyID
-      visibility: 'show' | 'hide'
+      visibility: 'show' | 'hide' | 'hide_if_empty'
     }>
   }
 }


packages/notion-types/src/collection.ts
```diff
@@ -46,7 +46,7 @@ export interface Collection {
     }>
     property_visibility?: Array<{
       property: PropertyID
-      visibility: 'show' | 'hide'
+      visibility: 'show' | 'hide' | 'hide_if_empty'
```
normdoow, line 49: Hey @nabettu,
Adding the def type doesn't actually hide the properties on the page. Is this supposed to make that change?

nabettu: oh Sorry I add only type.
I will change collection-row.tsx for this property. plz wait.

PR #78, opened 2021-06-01
Added captions for code blocks

normdoow:
Here is a notion page id to test on: 54ffff21b193431f98651dc2d776ab34
Just adds the caption to the code block if there is one.

image
The text right underneath the block is a caption in this image.


transitive-bullshit:
Thanks, Noah 🙏 😊

transitive-bullshit:
Published as `v4.6.4`

PR #86, opened 2021-06-08
Fixed board collection in toggle causing crash

normdoow:
Example of this issue on this page id: 7d3913ebc76a4ac1abc20e5a12bec55a

When the board is in a toggle the `groupResults` data is not present.
I couldn't find it anywhere else in the block either. So we aren't able to render the table when its inside a toggle.

The data without groupResults
Screen Shot 2021-06-08 at 12 35 15 PM
The data with groupResults
Screen Shot 2021-06-08 at 12 35 47 PM

May need extra 👀 on this. But I'm not seeing away we can render the table in this case.

So for now I am return null if the `groupResults` is not there.


transitive-bullshit:
I think this is fine; I don't see any possible cases where more checks could hurt anything.

transitive-bullshit:
Published as `v4.7.1`.

PR #87, opened 2021-06-11
Add support for hideBlockId param on NotionRenderer

### LINKED ISSUE #84: Is the HTML class including the Notion page ID necessary?
I'm exploring whether it's possible to provide a Notion sharing service that obfuscates the original Notion page IDs to a reasonable extent - reducing the impact of the page needing to be made public (until we can use the official API). It looks like the only place the Notion page ID slips into the rendered content is the top-level div, which is eg. `notion notion-app light-mode notion-block-067dd719a912471ea9a3ac10710e7fdf`. I can't see any CSS or anything using the class `notion-block-067dd719a912471ea9a3ac10710e7fdf`.

Is this class with the page ID necessary, and if not then would you accept a PR that either removed it or added a prop to do so?

mcintyre94:
Closes #84 

This PR adds an optional new parameter `hideBlockId` to the `NotionRenderer`:

```javascript
<NotionRenderer recordMap={recordMap} fullPage={true} darkMode={false} hideBlockId={true} />
```

If set to true then the `blockId` used to generate HTML classes is hardcoded to `notion-block`, instead of appending the Notion block ID. This allows a (server rendered) library user to keep the Notion page block ID private, which reduces the downside of needing to make a page public while react-notion-x can't use the official API.

If not passed (ie by default) block IDs remain unchanged, the change is entirely opt-in.

### DIFF
--- a/packages/react-notion-x/src/block.tsx
+++ b/packages/react-notion-x/src/block.tsx
@@ -37,6 +37,8 @@ interface BlockProps {
   pageFooter?: React.ReactNode
   pageAside?: React.ReactNode
   pageCover?: React.ReactNode
+
+  hideBlockId?: boolean
 }
 
 const tocIndentLevelCache: {
@@ -68,7 +70,8 @@ export const Block: React.FC<BlockProps> = (props) => {
     pageHeader,
     pageFooter,
     pageAside,
-    pageCover
+    pageCover,
+    hideBlockId
   } = props
 
   if (!block) {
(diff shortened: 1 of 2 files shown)


transitive-bullshit:
This looks perfect & well thought-out.

Much appreciated, @mcintyre94 🙏

transitive-bullshit:
Released as `v4.7.0`.

mcintyre94:
Thanks for getting this in so quickly @transitive-bullshit! :)

PR #88, opened 2021-06-12
Allow SSR for GracefulImage

mcintyre94:
This PR updates `GracefulImage` so that when rendered in a non-browser environment it falls back to an `` element instead of using react-image's ``. This allows static server rendering to correctly render images. Note that react-image deliberately does not SSR images: https://github.com/mbrevda/react-image/issues/206 (it renders nothing if only the src is provided which is how react-notion-x uses it)

In a browser the behaviour is unchanged. The props taken by `<GracefulImage` are unchanged.

Note that for simplicity the props are used as-is for the `` element, and I've used a ts-ignore for this. Basically if rendering `<GracefulImage>` on the server, it acts as a plain `` and only props for that tag will be relevant. I don't think it's worth doing anything more sophisticated than this yet, because all current use cases use only ``-friendly tags. 

This fixes missing page + header icons when server rendering.

| Before | After |
|--------|-------|
| Screenshot 2021-06-12 at 21 16 02    | Screenshot 2021-06-12 at 21 18 30   |

### DIFF
--- a/packages/react-notion-x/src/components/graceful-image.tsx
+++ b/packages/react-notion-x/src/components/graceful-image.tsx
@@ -1,3 +1,12 @@
-import { Img } from 'react-image'
+import React from 'react'
+import { Img, ImgProps } from 'react-image'
+import { isBrowser } from '../utils'
 
-export const GracefulImage = Img
+export const GracefulImage = (props: ImgProps) => {
+  if(isBrowser) {
+    return <Img {...props} />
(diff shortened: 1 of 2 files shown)


transitive-bullshit:
Another excellent PR; thanks, @mcintyre94 🙏

transitive-bullshit:
Published as `v4.7.2`

PR #91, opened 2021-06-18
Added support for sync blocks

normdoow:
The new Notion sync block is here. Now react-notion-x can support this.

I exported the NotionBlockRenderer sot that functionality could be used for the reference blocks.

Test page on: 8bcd65801a5d450fb7218d8890a38c29

### DIFF
--- a/packages/react-notion-x/src/components/sync-pointer-block.tsx
+++ b/packages/react-notion-x/src/components/sync-pointer-block.tsx
@@ -1,0 +1,35 @@
+import {
+  SyncPointerBlock as SyncPointerBlockType,
+  Block as BlockType
+} from 'notion-types'
+import React from 'react'
+
+import { NotionBlockRenderer } from '../renderer'
+
+export const SyncPointerBlock: React.FC<{
+  block: BlockType
+  level: number
+}> = (props) => {
+  let { block, level } = props
+
+  if (!block) {
+    if (process.env.NODE_ENV !== 'production') {
+      console.warn('missing block', block.id)
+    }
+    return null
+  }
+  const syncPointerBlock = block as SyncPointerBlockType
+  const referencePointerId =
+    syncPointerBlock?.format?.transclusion_reference_pointer.id
+  if (referencePointerId == null) return
+
+  if (referencePointerId == null) return null
+
+  return (
+    <NotionBlockRenderer
+      key={referencePointerId}
+      level={level}
+      blockId={referencePointerId}
+    />
+  )
+}
--- a/packages/react-notion-x/src/components/sync-pointer-block.tsx
+++ b/packages/react-notion-x/src/components/sync-pointer-block.tsx
@@ -10,7 +10,7 @@ export const SyncPointerBlock: React.FC<{
   block: BlockType
   level: number
 }> = (props) => {
-  let { block, level } = props
+  const { block, level } = props
 
   if (!block) {
     if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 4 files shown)


packages/react-notion-x/src/components/sync-pointer-block.tsx
```diff
@@ -0,0 +1,35 @@
...
+} from 'notion-types'
+import React from 'react'
+
+import { NotionBlockRenderer } from '../renderer'
+
+export const SyncPointerBlock: React.FC<{
+  block: BlockType
+  level: number
+}> = (props) => {
+  let { block, level } = props
```
transitive-bullshit, line 13: const?

packages/react-notion-x/src/components/sync-pointer-block.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  if (!block) {
+    if (process.env.NODE_ENV !== 'production') {
+      console.warn('missing block', block.id)
+    }
+    return null
+  }
+  const syncPointerBlock = block as SyncPointerBlockType
+  const referencePointerId =
+    syncPointerBlock?.format?.transclusion_reference_pointer.id
```
transitive-bullshit, line 23: better to just make it all `?.` just in case

packages/react-notion-x/src/components/sync-pointer-block.tsx
```diff
@@ -0,0 +1,35 @@
...
+  if (referencePointerId == null) return
```
transitive-bullshit, line 24: remove this line

packages/react-notion-x/src/components/sync-pointer-block.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  if (referencePointerId == null) return null
```
transitive-bullshit, line 26: instead of checking for null, just check:

`if (!referencePointerId)`

when you do `?.` if the latter part doesn't exist, it'll return undefined.

in general in js/ts, it's more robust to try and check for falsiness instead of specific values like this

packages/react-notion-x/src/components/sync-pointer-block.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  return (
+    <NotionBlockRenderer
```
transitive-bullshit, line 29: i think `hideBlockId` won't get passed down properly in this case.. it should probably be part of the context that's setup when we initialize `NotionContextProvider`

transitive-bullshit: this is a minor thing though and i don't think it should block this PR

transitive-bullshit (changes requested):
looks really great, noah 🙏 

just a few small comments

packages/react-notion-x/src/components/sync-pointer-block.tsx
normdoow: Great point!

packages/react-notion-x/src/components/sync-pointer-block.tsx
normdoow: And what's the hideBlockId do?

normdoow:
@transitive-bullshit Thanks for the comments. 👍  Made those changes.

packages/react-notion-x/src/components/sync-pointer-block.tsx
normdoow: yeah, this wasn't supposed to be there. 😅

packages/react-notion-x/src/components/sync-pointer-block.tsx
mcintyre94: @normdoow I added that in https://github.com/NotionX/react-notion-x/pull/87 - it removes the block ID from HTML classes

normdoow: ahh, got it.

transitive-bullshit (approved):

PR #95, opened 2021-06-22
Fixed search not having the right recordmap

normdoow:
Testing on this public notion id: e8da438394ba45f19ff4469b239e23ab

Here is the problem. mapPageUrl doesn't have the correct recordMap for all the search pages. So then you can't create the correct url for search results using automatic urls for example.

So when searching you get this:

Now you can pass the results recordMap in and have more info to create the results needed for mapPageUrl. So you get this now.

You could create a wrapper around this to pass a generic mapPageUrl into the NotionRenderer. But that recordMap only has the pages in the recordmap for that page. It doesn't have all the nested pages of a multilayer document.

This is also backwards compatible as recordMap can be undefined.

### DIFF
--- a/packages/react-notion-x/src/components/search-dialog.tsx
+++ b/packages/react-notion-x/src/components/search-dialog.tsx
@@ -98,7 +98,7 @@ export class SearchDialog extends React.Component<{
                             <components.pageLink
                               key={result.id}
                               className={cs('result', 'notion-page-link')}
-                              href={mapPageUrl(result.block.id)}
+                              href={mapPageUrl(result.block.id, searchResult.recordMap)}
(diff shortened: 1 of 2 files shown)


normdoow:
Hope this issue makes sense @transitive-bullshit.

MichaelDevC:
I'm having the same 'issue', I try to make a nested page slug system, everything works very well but the search function not for that reason... Hope in future will be implemented that fix

normdoow:
> I'm having the same 'issue', I try to make a nested page slug system, everything works very well but the search function not for that reason... Hope in future will be implemented that fix

This  should fix it.

PR #100, opened 2021-07-02
fixed multiple react apps

dialegein:
https://github.com/NotionX/react-notion-x/issues/85#issuecomment-860096671


transitive-bullshit:
Thanks @dialegein 🙏

PR #101, opened 2021-07-02
fixed TOC display in columns

dialegein:
pageBlock.value.content  does not contain nested blocks
https://github.com/craigary/nobelium/issues/138#issue-907925850

### DIFF
--- a/packages/notion-utils/src/get-page-table-of-contents.ts
+++ b/packages/notion-utils/src/get-page-table-of-contents.ts
@@ -19,10 +19,9 @@ const indentLevels = {
  * H1, H2, and H3 elements.
  */
 export const getPageTableOfContents = (
-  page: types.PageBlock,
   recordMap: types.ExtendedRecordMap
 ): Array<TableOfContentsEntry> => {
-  const toc = (page.content ?? [])
+  const toc = (Object.keys(recordMap?.block) ?? [])
     .map((blockId: string) => {
       const block = recordMap.block[blockId]?.value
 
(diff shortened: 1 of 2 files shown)


normdoow:
@dialegein Can you share a notion page Id that has an example of this issue that we can test?

dialegein:
> @dialegein Can you share a notion page Id that has an example of this issue that we can test?

1173a216ce144410b2796955dbb9b1bb

normdoow:
@dialegein so currently the TOC only appears if the header is in the first column correct? But not if it is in a secondary collumn.
The images from the issue are a little confusing as to which usecase they are testing.

dialegein:
> @dialegein so currently the TOC only appears if the header is in the first column correct? But not if it is in a secondary collumn.
> The images from the issue are a little confusing as to which usecase they are testing.

the issue is that  the TOC may lose some headers, if header in columns. whether it’s the first or second column.sorry my English is poor so I may not explain clearly

normdoow (approved):
Looks good!

normdoow:
Live in version 4.7.8!

transitive-bullshit:
Hmmm, the order of the blocks in the recordMap doesn't necessarily match the order of the blocks in the page... I'm not sure if it's safe to rely on this?

normdoow:
> Hmmm, the order of the blocks in the recordMap doesn't necessarily match the order of the blocks in the page... I'm not sure if it's safe to rely on this?

Ah, I did not think of that. Though I haven't seen too many weird issues with that so far with the sites I've seen.

PR #107, opened 2021-07-22
Fix for board collections

normdoow:
Testing on: https://www.notion.so/potionsite/newest-board-a899b98b7cdc424585e5ddebbdae60cc

Issue: https://github.com/NotionX/react-notion-x/issues/104

Looks like Notion changed the name in collectionview.format from board_groups2 to board_columns.
We grab the queryCollection data and add it on to the recordMap. That notion endpoint was failing because it didn't have the right query params with the name change.


normdoow:
Updated so that the query with group_by is set correctly.
This is also backwards compatible.

PR #108, opened 2021-07-26
Check for undefined in block property

mannyyang:
Use a conditional on top of the nullish coalescing operator

### DIFF
--- a/packages/notion-utils/src/get-page-table-of-contents.ts
+++ b/packages/notion-utils/src/get-page-table-of-contents.ts
@@ -21,14 +21,9 @@ const indentLevels = {
 export const getPageTableOfContents = (
   recordMap: types.ExtendedRecordMap
 ): Array<TableOfContentsEntry> => {
-  let recordMapBlock: any = []
-  if (recordMap && recordMap.block) {
-    recordMapBlock = recordMap.block
-  }
-  
-  const toc = (Object.keys(recordMapBlock) ?? [])
+  const toc = (Object.keys(recordMap?.block ?? {}) ?? [])
     .map((blockId: string) => {
-      const block = recordMapBlock[blockId]?.value
+      const block = recordMap.block[blockId]?.value
 
       if (block) {
         const { type } = block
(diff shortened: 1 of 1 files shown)


packages/notion-utils/src/get-page-table-of-contents.ts
```diff
@@ -21,9 +21,14 @@ const indentLevels = {
...
-  const toc = (Object.keys(recordMap?.block) ?? [])
+  let recordMapBlock: any = []
+  if (recordMap && recordMap.block) {
+    recordMapBlock = recordMap.block
+  }
+  
+  const toc = (Object.keys(recordMapBlock) ?? [])
     .map((blockId: string) => {
-      const block = recordMap.block[blockId]?.value
+      const block = recordMapBlock[blockId]?.value
```
normdoow, line 31: Doesn't the coalescing operator already handle this? the recordmap array  will just be null if it is null.

mannyyang: yeah that was my understanding as well, but maybe it doesn't compile correctly on production? Here's a live example of the issue. It seems to flash the content correctly then shows the error page.

https://www.notionpotions.space/
Error:
image


here's how the compiled code looks like:
image

To me it looks like it checks to see whether `e` (or recordMap) is available and if so run `Object.keys(e.block)`. `e.block` can be undefined which will throw the error since Object.keys requires an array. Also looking at it again, it seems like if recordMap is undefined, it will throw an error as well.

mannyyang: A better solution could be `(Object.keys(recordMap?.block ?? []) ?? [])`

mannyyang:
@normdoow a heads up I updated it to use another coalescing operator.

normdoow:
> @normdoow a heads up I updated it to use another coalescing operator.

Looks better. But curious how we are getting to this point in the first place where recordmap is null in the first place. Recordmap shouldn't be null. the function doesn't accept null.

normdoow:
> > @normdoow a heads up I updated it to use another coalescing operator.
> 
> Looks better. But curious how we are getting to this point in the first place where recordmap is null. Recordmap shouldn't be null. the function doesn't accept null.

ahh sorry. I see the block can be.

normdoow (approved):

normdoow:
@mannyyang live on v4.8.1

mannyyang:
Thank you!!

PR #114, opened 2021-08-04
Support for new Notion callout blocks

MichaelDevC:
image

Test Page: https://michaeldevc.notion.site/Testing-new-callout-blocks-3229861ab6254424bf0c8fd4c1fc9dc8

### DIFF
--- a/packages/react-notion-x/src/block.tsx
+++ b/packages/react-notion-x/src/block.tsx
@@ -652,13 +652,7 @@ export const Block: React.FC<BlockProps> = (props) => {
 
           <div className='notion-callout-text'>
             <Text value={block.properties?.title} block={block} />
-            <>
-              {
-                block.content?.map((elementId) => (
-                  <Block block={recordMap.block[elementId].value} level={1} />
-                ))
-              }
-            </>
+            {children}
           </div>
         </div>
       )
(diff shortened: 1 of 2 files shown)


packages/react-notion-x/src/block.tsx
```diff
@@ -652,6 +652,13 @@ export const Block: React.FC<BlockProps> = (props) => {
 
           <div className='notion-callout-text'>
             <Text value={block.properties?.title} block={block} />
+            <>
+              {
+                block.content?.map((elementId) => (
+                  <Block block={recordMap.block[elementId].value} level={1} />
+                ))
+              }
+            </>
```
normdoow, line 661: This can actually be
```
<div className='notion-callout-text'>
  <Text value={block.properties?.title} block={block} />
  {children}
</div>
```
Because the NotionRenderer which is the parent already renders block content as children.
This is how the sync block works as well if you want to take a look at that for example.

normdoow (changes requested):
Tested and it looks good. But could you make the one change I mentioned?

Thanks!

packages/react-notion-x/src/block.tsx
MichaelDevC: Oh, that's right. Just updated!

normdoow (approved):

normdoow:
live on version 4.8.3 now!

PR #119, opened 2021-08-12
feat: added custom checkboxes

### LINKED ISSUE #118: Support interactive checkboxes
Introducing a way to add custom checkboxes to the notion renderer. The objective is to have interactive checkboxes for the visitors which the app would link as a controlled input. 

I think the changes would be made on [this](https://github.com/NotionX/react-notion-x/blob/d4d17e696a9f4777c1c96ab5b8f29aea1e4e9800/packages/react-notion-x/src/block.tsx#L785) line 

If the maintainers are ok with the change, I can go ahead and pick this up as well.

Utkarshbhimte:
resolves #118

### DIFF
--- a/packages/react-notion-x/src/block.tsx
+++ b/packages/react-notion-x/src/block.tsx
@@ -782,7 +782,10 @@ export const Block: React.FC<BlockProps> = (props) => {
       return (
         <div className={cs('notion-to-do', blockId)}>
           <div className='notion-to-do-item'>
-            <Checkbox isChecked={isChecked} />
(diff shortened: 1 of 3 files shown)


normdoow:
Hey @Utkarshbhimte.
There were a couple of issues with this here: https://github.com/NotionX/react-notion-x/pull/129.
Take a look and maybe that will help for when you make a similar change in the future.
Also, be sure to run to see if there are any type errors or similar.

normdoow:
This is now live on v4.8.8!

PR #132, opened 2021-09-17
Fix image caption and make images with url captions a link

normdoow:
Testin on: 3492bd6dbaf44fe7a5cac62c5d402f06

Made captions aligned with images. Before they were right aligned even when the image is centered
image

Made images like this one clickable with a link.
To do this set the caption to be just a valid url or a notion page id.
like https://google.com or /3492bd6dbaf44fe7a5cac62c5d402f06
image


normdoow:
pushed to v4.9.0

PR #134, opened 2021-09-20
Use signed urls in images that needs them

### LINKED ISSUE #68: Uploaded images are not rendering
Hi, I am trying to use this package, but for some reason uploaded images look broken. Unsplash images and images added through links are working ok.

Screenshot 2021-05-16 at 10 12 02 AM

Here when I right-click on the image and select `Open image in new tab`, the image is properly loading, but it is not being rendered on the actual page. Not sure if I am doing something wrong here. Any help with this would be very helpful.

These are the props that I added. Do I have to add anything else extra to these?
```jsx
 <NotionRenderer
  recordMap={extendedRecordMap}
  components={{
    code: Code,
    collection: Collection,
    collectionRow: CollectionRow,
    tweet: Tweet,
    pdf: Pdf,
    equation: Equation,
  }}
/>
```

remorses:
Some images seem to require signedUrl as src to add some parameters like `X-Amz-Credential`

An example of a page that requires these signed urls is this one https://brave-iberis-6ea.notion.site/The-basics-creating-updating-managing-issues-da8e0c324b784bc48b3797aa9753a198

Should fix #68

### DIFF
--- a/packages/notion-client/src/notion-api.ts
+++ b/packages/notion-client/src/notion-api.ts
@@ -185,7 +185,7 @@ export class NotionAPI {
           block &&
           (block.type === 'pdf' ||
             block.type === 'audio' ||
-            block.type === 'image' ||
+            (block.type === 'image' && block.file_ids?.length) ||
             block.type === 'video' ||
             block.type === 'file')
         ) {


normdoow:
Hey thanks for this pull request! I'll take a look.

PR #137, opened 2021-09-30
style: :art: change position in .notion-viewport to relative

yehezkielgunawan:
Related to this issue [https://github.com/NotionX/react-notion-x/issues/135](https://github.com/NotionX/react-notion-x/issues/135). 

Found that in the class `notion-viewport` there's a CSS position `position: fixed` and that makes the overlay on the page that makes the user cannot interact with the page.

### DIFF
--- a/packages/react-notion-x/src/styles.css
+++ b/packages/react-notion-x/src/styles.css
@@ -231,7 +231,7 @@
 }
 
 .notion-viewport {
-  position: fixed;
+  position: relative;
   top: 0;
   left: 0;
   right: 0;


normdoow:
I don't think we necessarily want to do this. making it relative basically removes the viewport completely. I haven't been able to repro the issue of not being able to click through the content. Can you share more how you were seeing that?

d-ivashchuk:
Just using the notion renderer and not using `fullPage` should be enough to reproduce. If not I could try give you a sandbox!

yehezkielgunawan:
Yes @d-ivashchuk right, when just using the notion renderer just to render the text from the notion page (not using `fullPage`), this issue happens.

normdoow:
Ok, I'll try to get to this. Have some other things on my plate currently. Hopefully your workaround can work till then!

futantan:
> I still wasn't able to repro this issue. have a fixed notion-viewport with fullPage: false. I'm able to click through to other pages.

Just change the `fullPage` props to `false` can reproduce the issue.

Here is the demo:
CleanShot 2022-01-18 at 23 17 46

I can select nothing but the links.

futantan:
It seems that `notion-viewport` is there for image zooming, but the `zoomable` is set to false.

@normdoow Can you provide more info so that maybe I can help with that.

normdoow:
> It seems that `notion-viewport` is there for image zooming, but the `zoomable` is set to false.
> 
> @normdoow Can you provide more info so that maybe I can help with that.

Ahh yes. We used to have the zoomable so that users could zoom in on images. But we moved away from that being the default as it isn't used for most usecases.

futantan:
> > It seems that `notion-viewport` is there for image zooming, but the `zoomable` is set to false.
> 
> > 
> 
> > @normdoow Can you provide more info so that maybe I can help with that.
> 
> 
> 
> Ahh yes. We used to have the zoomable so that users could zoom in on images. But we moved away from that being the default as it isn't used for most usecases.

Back to this issue, are we safe to remove notion-viewport div for now?

normdoow:
> 

Great idea. We are actually slowly doing this! just added a custom component for image recently. some others have this already as well. Problem for the zoomable part of the image is we will still need this viewport around.

normdoow:
> > > It seems that `notion-viewport` is there for image zooming, but the `zoomable` is set to false.
> > 
> > 
> > > 
> > 
> > 
> > > @normdoow Can you provide more info so that maybe I can help with that.
> > 
> > 
> > Ahh yes. We used to have the zoomable so that users could zoom in on images. But we moved away from that being the default as it isn't used for most usecases.
> 
> Back to this issue, are we safe to remove notion-viewport div for now?

I guess that makes sense to remove the viewport for now. and can add it back later once we get the zoomable image figured out.

transitive-bullshit:
For some background context, the only reason notion-viewport originally existed is because I was trying to keep the CSS as close to notion's structure as possible. I'm not sure if this is still the case, and I always thought the previous `position: fixed` seemed hacky.

normdoow:
Thanks @transitive-bullshit. Maybe more we can do with this in the future. But this is pushed in v4.13.4 for now.

PR #140, opened 2021-10-06
fixed notion api query collection hitting 400 error

normdoow:
Fixes all tables except for board tables. They have some extra parameters that need sent. Will have to fix that later as a followup.


normdoow:
live now in v4.9.2

PR #143, opened 2021-10-07
fix: board collection view

pbteja1998:


normdoow:
Live on v4.9.5

normdoow:
Thanks @pbteja1998!

PR #149, opened 2021-10-14
Add addSignedUrls method

remorses:
Refactored the logic to add signed urls to record map in a separate method, this allows you to sign urls on an existing record map

This is useful because signed urls expire after a few days and if you store the record map in a database you need to refresh them before serving them to the user


remorses:
It's ready

Issue #26, opened 2021-02-07
Add support for using next/image


theodorepauw:
Would it be possible to enable [Next's Image component](https://nextjs.org/docs/api-reference/next/image)?

This would bring [built-in optimization](https://nextjs.org/docs/basic-features/image-optimization) without an external CDN.

transitive-bullshit:
Hey @theodorepauw, this is definitely on my bucket list, though it's not as easy as you would think. The main reason is that `next/image` currently isn't a drop-in replacement for `<img>` because it doesn't support most CSS styling.

The main difficulty is in making it possible to switch between different `Image` components and having CSS styling work gracefully between all of them, and with `next/image` containing multiple sub divs, this is difficult to do in practice.

Two related `next/image` issues are: https://github.com/vercel/next.js/issues/18858 and https://github.com/vercel/next.js/issues/18585, both of which I've commented on.

theodorepauw:
I understand. Thanks for the swift response!

TheMrZZ:
Hi!
NextJS implemented all necessary properties - would using `components.Image` instead of `GracefulImage` be possible now?

transitive-bullshit:
Yep, this was added in #158 

We still need to document and add an example of this, though. Just set `components.Image` to `Image` from `next/image`.

TheMrZZ:
It's not working - I already had `components.Image` set to `Image` from `next/image`, but it does not wrap any Notion image.

It's easy to notice: all `next/image` URLs start with `/_next/image?url=`, yet no image rendered by `react-notion-x` starts with it.

eagor:
see https://github.com/NotionX/react-notion-x/issues/203#issuecomment-1041971729

PR #158, opened 2021-11-05
Custom image component - next/image

MichaelCasaDev:
- Using next/image instead of custom LazyImage to render page images.
- With this update now we can use the Vercel image optimization with react-notion-x!! :fire:
- Updated example config to use next/image

_Tested via the example in the project repo._

Thanks to the update of Next.JS 12 I used the next/image component to ensure a better user experience, for image loading, using the Vercel Image Optimization.

Open to any suggestions.


normdoow:
@MichaelCasaDev this is awesome! I'll take a look and test. This definitely a big upgrade I have been wanting to do.
I think we are going to want to add a flag where people can turn this on or not. Vercel Image Optimization may not work for everyone since there are some # of image limits to it.

packages/react-notion-x/package.json
```diff
@@ -30,7 +30,8 @@
     "react-lazy-images": "^1.1.0",
     "react-modal": "^3.14.3",
     "react-pdf": "^5.4.1",
-    "react-use": "^17.3.1"
+    "react-use": "^17.3.1",
+    "next": "^12.0.2"
```
transitive-bullshit, line 34: `next` shouldn't be a dependency of `react-notion-x`.

Currently, `react-notion-x` isn't tied to next.js (it can be used with any react-based web framework).

If we were to tie it to next.js, it still wouldn't be a dependency of `react-notion-x`; it would be a peer dep or an optional dep.

packages/react-notion-x/src/block.tsx
```diff
@@ -531,20 +531,20 @@ export const Block: React.FC<BlockProps> = (props) => {
     case 'excalidraw':
     // fallthrough
     case 'image':
-    // fallthrough
+      return <AssetWrapper blockId={blockId} block={block} darkMode={darkMode} />
```
transitive-bullshit, line 534: No need to pass down `darkMode` as it's available from the react context.

packages/react-notion-x/src/components/asset.tsx
```diff
@@ -3,7 +3,7 @@ import { BaseContentBlock, Block } from 'notion-types'
 import { getTextContent } from 'notion-utils'
 
 import { useNotionContext } from '../context'
-import { LazyImage } from './lazy-image'
+import NextImage from 'next/image'
```
transitive-bullshit, line 6: We will likely need to add `image` to our overridable props s.t. the client application inserts `next/image` instead of the library itself. This will allow us to not take `next` as a dependency.

packages/react-notion-x/src/components/asset.tsx
```diff
@@ -217,13 +212,20 @@ export const Asset: React.FC<{
     const caption = getTextContent(block.properties?.caption)
     const alt = caption || 'notion image'
 
+    // Generate RGB-based base64 images for image placeholder while loading
+    const keyStr = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/='
```
transitive-bullshit, line 216: L215-L218 don't belong in a production library.

The current way the preview image is handled is by using `recordMap.preview_images` to store a map of image source URLs to preview image URLs.

transitive-bullshit (changes requested):
Thanks for getting the ball rolling on this @MichaelCasaDev. I agree that this would be a great improvement as discussed in #26.

With that being said, this PR would require major changes in order to use. See comments for details.

- next.js must not be a dependency of `react-notion-x`
- should pass all styling tests on `3492bd6dbaf44fe7a5cac62c5d402f06` 
- should allow for custom preview images (as the current solution allows via `recordMap.preview_images`)
- should support image zoom (via the `medium-zoom` package exposed via `useNotionContext`)

To be clear, I agree that this is the direction the project should go and really appreciate this PR 😃 Just want to make sure that it's done the right way. The original LazyImage impl was pretty hacky and `next/image` would be preferred for 95% of use cases, but I'd also like to make sure that non-next.js use cases are covered.

Thanks again @MichaelCasaDev 💪

packages/react-notion-x/package.json
transitive-bullshit: for an example of how to do this, see how we're handling `next/link` via the custom `pageLink` prop so that it's not part of the package itself but rather overridable by the client. https://github.com/transitive-bullshit/nextjs-notion-starter-kit/blob/main/components/NotionPage.tsx#L228

packages/react-notion-x/src/block.tsx
MichaelCasaDev: I removed the return of the AssetWrapper and fall through everything.
I also removed the "darkMode" and the new "vercelImages" and used the notion context.

packages/react-notion-x/src/components/asset.tsx
MichaelCasaDev: I removed `next` as dependency! 695690f

packages/react-notion-x/package.json
MichaelCasaDev: I made a new custom prop `image` to use custom images components, like next/image! 695690f

MichaelCasaDev: I'm still working on how to perfect the medium-zoom component to work properly with next/image

packages/react-notion-x/src/components/asset.tsx
MichaelCasaDev: Just made a new commit and removed the preview placeholder! d235b40

packages/react-notion-x/package.json
MichaelCasaDev: Just pushed a new commit to allow custom preview images! 3345717

example/pages/[pageId].tsx
```diff
@@ -72,8 +73,28 @@ export default function NotionPage({ recordMap }) {
...
+          }) => (
+            <Image
+              src={src}
+              alt={alt}
+              height={height}
+              width={width}
+              objectFit={objectFit}
+              unoptimized={unoptimized}
+              onLoad={onLoad}
+            />
```
Comment (iuzn) [inline example/pages/[pageId].tsx:96]: Notion images has a **"&width="** query for resizing. A similar method can be used for both responsive images and blurred placeholders.

```js
// Loader
const notionImageLoader = ({ src, width }) => {
  return `${src}&width=${width}`
}

// Image component
<Image
  loader={notionImageLoader}
  src={src}
  alt={alt}
  height={height}
  width={width}
  placeholder="blur"
  blurDataURL={`${src}&width=16`}
  ...
/>
```

Comment (normdoow) [reply example/pages/[pageId].tsx:96]: Yeah, this could be really nice to use. We may not need Vercel image optimization then?

Comment (MichaelCasaDev) [reply example/pages/[pageId].tsx:96]: I think as @normdoow says, it could be a really nice feature but I think we can let the user chose if use a loader or not.

normdoow:
@MichaelCasaDev where are we at with this?

MichaelCasaDev:
> @MichaelCasaDev where are we at with this?

I'm almost done, the last thing that I've to do is to fix the size of the images because now it's doesn't fit the styling tests.

frankcbliu:
> > @MichaelCasaDev where are we at with this?
> 
> I'm almost done, the last thing that I've to do is to fix the size of the images because now it's doesn't fit the styling tests.

good job! waiting for it.

pbteja1998:
This is awesome. Having to provide my own `Image` Component would be so cool. 

I can do all the image optimizations from my side itself, like including the `srcsets` for example.

Really looking forward to this PR being merged.

Also, once this is available, I don't think I will be using Vercel's image optimization or `next/image` for that matter. I would probably use a normal `` tag and add the relevant `srcsets` with various image sizes. 

`next/image` adds so many extra `divs` to the layout which I don't like. I would rather have something that is a drop-in replacement for normal `` tag.

MichaelCasaDev:
Almost everything is done! 🎉 

In the latest commits, I added the support for:
- Flag to enable or disable easily custom images ( `customImages` )
- Preview images support with a custom component
- Custom image component with lot of keys ( `src - alt - className - style - loading - decoding - width - height - ref - onLoad` ) **!! _height_ and _width_ are only provided with preview images !!**
- Support for zoomable images ( `medium-zoom` package )
- This PR doesn't use _next_ as a dependency in the project so the user can run this package also on plain react projects

pbteja1998:
@MichaelCasaDev what do you mean by `height and width are only provided with preview images`?

MichaelCasaDev:
> @MichaelCasaDev what do you mean by `height and width are only provided with preview images`?

So we can't provide a valid height and width on all images, because images are fetched during the render, and not before. 
For preview images, we already know the exact sizes of them.

When preview images are "disabled" the renderer returns a null value for width and height because we can't know them until the image is fetched and rendered.

normdoow:
@MichaelCasaDev are you thinking this PR is done? If so I can review.

MichaelCasaDev:
> @MichaelCasaDev are you thinking this PR is done? If so I can review.

Yes, I think we can review it now.

I tested it, and all functions work (preview images, full-width, and zoom)!

example/package.json
```diff
@@ -12,7 +12,7 @@
     "deploy": "vercel --prod"
   },
   "dependencies": {
-    "next": "^11.1.2",
+    "next": "^12.0.3",
```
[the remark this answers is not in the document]
MichaelCasaDev: Just committed the update!

normdoow (approved):
Looks good. Thanks Michael!

normdoow:
live on v4.13.1!

PR #159, opened 2021-11-08
fix: NotionAPI collectionQuery error Cannot read property 'type' of…

Dionid:
Have the same error like https://github.com/NotionX/react-notion-x/issues/147 and found out that this `?` is crusial


transitive-bullshit:
Thanks @Dionid 💪 😄

PR #161, opened 2021-11-13
feat: Add support for custom callout blocks

CryogenicPlanet:
Hey, this is a small PR to add support for custom `callout` blocks. IMO this should be an option for all types of blocks, so that it can allow the user to customize the block however they want.


normdoow:
I agree with you that it should be an option for all blocks.

normdoow (approved):

normdoow:
published to v4.11.2 👌
Thanks @CryogenicPlanet.
If you want to add custom blocks for all types that would be cool!

Issue #163, opened 2021-11-15
notion-client - private page with page header image doesn't get the signed URL


marekhvolka:
When we have a private page with an image in the page header and this image was uploaded to Notion, we don't get a signed URL for it. Therefore the image can't be displayed

normdoow:
added to v4.11.4

PR #164, opened 2021-11-15
notion-client - return signed URL for page headers

marekhvolka:
This is a fix for [#163](https://github.com/NotionX/react-notion-x/issues/163)


normdoow:
This is looking good. Thanks for contributing!

normdoow:
added to v4.11.4

PR #169, opened 2021-12-03
fix: code block content lost

frankcbliu:
As the issues say: https://github.com/NotionX/react-notion-x/issues/168

here i fix it like that:
image


packages/react-notion-x/src/block.tsx
```diff
@@ -568,7 +572,7 @@ export const Block: React.FC<BlockProps> = (props) => {
 
     case 'code': {
       if (block.properties.title) {
-        const content = block.properties.title[0][0]
```
normdoow, line 571: Could you change it so there isn't a function call? I'd prefer it in one line.

normdoow:
Interesting bug. My code block only have one title block with code in it. Not sure what makes it so there are multiple. Maybe Notion updated that as well.

transitive-bullshit:
> Interesting bug. My code block only have one title block with code in it. Not sure what makes it so there are multiple. Maybe Notion updated that as well.

Every block has a version attached to it, so notion can internally make changes while remaining backwards compatible. I'd recommend that we try to always be defensive to try and support all block versions without having block version-specific logic (esp since it's really hard to test the diff btwn block versions).

packages/react-notion-x/src/block.tsx
transitive-bullshit: Could be useful to have a `getBlockTitle` function in `notion-utils` that does this in a safe way. Not necessary, just a thought.

normdoow:
> > Interesting bug. My code block only have one title block with code in it. Not sure what makes it so there are multiple. Maybe Notion updated that as well.
> 
> Every block has a version attached to it, so notion can internally make changes while remaining backwards compatible. I'd recommend that we try to always be defensive to try and support all block versions without having block version-specific logic (esp since it's really hard to test the diff btwn block versions).

This fix should do that pretty well then. As in work for all versions while not having to have version logic.

packages/react-notion-x/src/block.tsx
frankcbliu: ok, i will fix it.

frankcbliu: > 

you means create a function such as named `getCodeContent` to do this?

normdoow: `getBlockTitle` would probably take a block as input and do this `block.properties.title.map((arr) => { return arr[0] }).join('')`
probably need to use ?'s or have some checks that the title is there. But I'm guessing  thats what he means.
I'm fine either way.

frankcbliu: > getBlockTitle would probably take a block as input and do this block.properties.title.map((arr) => { return arr[0] }).join('')
probably need to use ?'s or have some checks that the title is there. But I'm guessing thats what he means.
I'm fine either way.

i got it. but i found that the function `getBlockTitle` will lost some **rich text formatting**, the code:
image

of course, generally speaking, there are no such `rich text characters` in the code, so the problem should not be big

frankcbliu:
maybe i broken the commit history, sorry for that, i will restart a new pull request.

PR #170, opened 2021-12-03
Made it so captions on embeds can be seen

normdoow:
There was a bug where no captions could be seen except for images.
Embed captions were hidden by the embed.

So I moved captions outside of the asset block for embeds. The images caption needs to stay inside the asset so it can have the correct styling when an image is centered.

PageId to test: 5d4e290ca4604d8fb809af806a6c1749

Fixed image example
Screen Shot 2021-12-03 at 2 08 03 PM

Fixed embed example
Screen Shot 2021-12-03 at 2 07 42 PM

### DIFF
--- a/packages/react-notion-x/src/components/asset.tsx
+++ b/packages/react-notion-x/src/components/asset.tsx
@@ -229,9 +229,12 @@ export const Asset: React.FC<{
   }
 
   return (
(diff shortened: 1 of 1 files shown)


transitive-bullshit:
Awesome 😄

PR #172, opened 2021-12-07
fix: align left for table of content

futantan:
image
This PR fix the styling above. 
Feel free to close this directly if you think the original one is better.

### DIFF
--- a/packages/react-notion-x/src/styles.css
+++ b/packages/react-notion-x/src/styles.css
@@ -181,7 +181,6 @@
 .notion-aside-table-of-contents {
   display: flex;
   flex-direction: column;
-  align-items: center;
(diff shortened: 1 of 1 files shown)


normdoow:
Can you link to a page where you are seeing this bad styling?
I'm not really seeing this on the page I'm looking at.
image
Table of contents seems to be algned correctly here on this pageId: 0c322c33381c49bca5083a451c334c39

futantan:
@normdoow 
It's here: https://transitivebullsh.it/nextjs-notion-starter-kit

normdoow:
Ahh I see. Looks like that is functionality specifically for the notion starter kit. I'll leave that decision up to @transitive-bullshit.I could see people wanting this either way.

Issue #176, opened 2021-12-10
Support for normalize chinese title


vv13:
When I execuse building script, I got some Collecting page data like this:
```
info  - Collecting page data ...[
  { params: { pageId: '500' } },
  { params: { pageId: '31cc504c078a4d23843fb9d0ae69c590' } },
]
```
and it throw some strange error:
```
> Build error occurred
[Error: ENOENT: no such file or directory, rename '/Users/xxx/nextjs-notion-starter-kit/.next/export/500.html' -> '/Users/xxx/nextjs-notion-starter-kit/.next/server/pages/500.html'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'rename',
  path: '/Users/xxx/nextjs-notion-starter-kit/.next/export/500.html',
  dest: '/Users/xxx/nextjs-notion-starter-kit/.next/server/pages/500.html'
}
```

In notion-utils package, I found in `src/get-canonical-page-id.ts `,  It including the function:
```
export const normalizeTitle = (title: string | null): string => {
  return (title || '')
    .replace(/ /g, '-')
    .replace(/[^a-zA-Z0-9-]/g, '')
    .replace(/--/g, '-')
    .replace(/-$/, '')
    .replace(/^-/, '')
    .trim()
    .toLowerCase()
}
```

Because of this, it converting my pageId from rawPageTitle `星空 500 法则` to `500`.

normdoow:
added to v4.11.4

PR #177, opened 2021-12-10
feat: support for normalize chinese title

vv13:
fix for issue #176 .

### DIFF
--- a/packages/notion-utils/src/get-canonical-page-id.ts
+++ b/packages/notion-utils/src/get-canonical-page-id.ts
@@ -33,7 +33,7 @@ export const getCanonicalPageId = (
 export const normalizeTitle = (title: string | null): string => {
   return (title || '')
     .replace(/ /g, '-')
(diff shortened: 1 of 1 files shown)


normdoow:
This looks good. Makes sense. One thing I'm not sure about. Do Chinese characters work as urls in all browsers?

vv13:
> This looks good. Makes sense. One thing I'm not sure about. Do Chinese characters work as urls in all browsers?

URL use [RFC-1738](https://www.ietf.org/rfc/rfc1738.txt), It basically support us-ascii sets, browsers will encode chinese charaters automatically, but there are differences between different browsers.

As far as i know,  some old browser like IE also support encode character to gb2312.

for this PR, It simply make `getCanonicalPageId` can also get chinese readable title, and if we use it as URL, almost all browser will automatically convert it to URL character set.

normdoow:
Looking good. thanks for this!

normdoow:
added to v4.11.4

Issue #168, opened 2021-12-03
[BUG] code block render error


frankcbliu:
Hello, you can see the content on my notion block here: https://frankcbliu.notion.site/32f6bb537afc49db909732e67023bbb3

we can see the code content is:
```
module.exports = {
  rootNotionPageId: '6a76fdd9ec6d4ae2855aa790d620eec4'
  ...
}
```
show on here:
![image](https://user-images.githubusercontent.com/36353894/144665770-a30fcda7-2210-4490-b87a-b760af835fe4.png)

but i got page from `getPage` API, the result is :
![image](https://user-images.githubusercontent.com/36353894/144671161-21c2b871-27dd-4726-94b2-1b14672507b1.png)

the notion-client retun's data is true, but the render result is error.

the render result is:
![image](https://user-images.githubusercontent.com/36353894/144666091-9ba103be-e925-4e42-8718-4511898c9bf4.png)

may be you can have a try. thanks.

frankcbliu:
sorry, it is my fault.
![image](https://user-images.githubusercontent.com/36353894/144666572-7e0c59fb-ed64-4ea2-8774-3e7b01eb67ac.png)

frankcbliu:
the notion-client retun's data is true, but the render result is error

normdoow:
added to v4.11.4

PR #178, opened 2021-12-10
fix: code block content lost

frankcbliu:
the old PR is here: #169 
the issue is here: #168


normdoow:
Looks good!

normdoow:
added to v4.11.4

PR #180, opened 2021-12-11
Support custom header

hcorta:
Hi! 
This PR extends @harajune proposal on [171](https://github.com/NotionX/react-notion-x/pull/171).
My main goal was enabling passing a custom header component, but allowing the information in the original to be accesible.

```
default function App() {

  function MyCustomHeader({
    activePageId,
    blockIds,
    blockMap,
    breadcrumbs,
    className,
    components,
    hasSearch,
    isSearchOpen,
    mapPageUrl,
    onCloseSearch,
    onOpenSearch,
    recordMap,
    rootPageId,
    searchNotion,
    setIsSearchOpen,
    ...rest
  }) {
    return <header>{'my custom header'}</header>
  }

  return (
      <NotionRenderer
         header={MyCustomHeader}
      />
  )
}
```

A few comments:
1. I have renamed the "page-header.tsx" to "header.tsx" (as well as the component) since the "pageHeader" prop refers to another component and it might leads to confusion. 
2. I have created an util for getting the breadcrumbs into the utils package -> "get-page-breadcrumbs.tsx"
3.  I have added the "disableHeader" prop to fully disable the header, if necessary.

As this is my first PR to the lib, I'd willing to hear suggestions on code style or similar.
🙂


packages/react-notion-x/src/components/header.tsx
```diff
@@ -0,0 +1,151 @@
+import React from 'react'
```
iNorthIsle, line 1: LGTM! But the PR will look better and less confused if change it in the `page-header` file while not creating a new one and deleting it 😁

normdoow:
Thanks @hcorta will be looking at this over the next couple of days.

packages/react-notion-x/src/block.tsx
```diff
@@ -32,13 +32,15 @@ interface BlockProps {
   className?: string
   bodyClassName?: string
 
+  header?: React.ElementType
   footer?: React.ReactNode
   pageHeader?: React.ReactNode
```
normdoow, line 37: aren't you replacing this pageHeader so it should be removed?

packages/react-notion-x/src/block.tsx
```diff
@@ -66,12 +68,14 @@ export const Block: React.FC<BlockProps> = (props) => {
     level,
     className,
     bodyClassName,
+    header,
     footer,
     pageHeader,
```
normdoow, line 73: aren't you replacing this pageHeader so it should be removed?

packages/react-notion-x/src/components/header.tsx
```diff
@@ -0,0 +1,150 @@
...
+import { getPageBreadcrumbs } from 'notion-utils'
+import { useNotionContext } from '../context'
+import { PageIcon } from './page-icon'
+import { SearchIcon } from '../icons/search-icon'
+import { cs } from '../utils'
+import { SearchDialog } from './search-dialog'
+
+export const Header: React.FC<{
+  header: React.ElementType
+}> = ({ header: CustomHeader }) => {
```
normdoow, line 13: Maybe I'm missing something but where is this CustomHeader type coming from? Obviously it requires some props.

normdoow:
Just a few comments! Thanks! Looks pretty good though!

hcorta:
@normdoow the pageHeader prop is not affected by my changes, and it is still is use, as it was, in **_<Block />_**. 
**_<CustomHeader/>_** is just the prop "**_header_**" that needs to be capitalised so it can to be used as a JSX element. I may rename it to <Header />, would that make it clearer?

normdoow:
> @normdoow the **_pageHeader_** prop is not affected somehow by my changes, and it is still is use, as it was, in **_Block_**. **_CustomHeader_** is just the prop "**_header_**" that needs to be capitalised so it can to be used as a JSX element. I may rename it to **_Header_**, would that make it clearer?

No, CustomHeader is good. I just didn't see it imported from anywhere.

hcorta:
@normdoow anything else here I can help with?

normdoow:
All good!

normdoow:
added to v4.12.0

pbteja1998:
@normdoow @hcorta `PageHeader` is something different and I am using it. Any reason why it got removed here?
CleanShot 2021-12-18 at 07 59 35@2x


Page header previously used to be on top of the page title.

normdoow:
> @normdoow @hcorta `PageHeader` is something different and I am using it. Any reason why it got removed here? CleanShot 2021-12-18 at 07 59 35@2x
> 
> Page header previously used to be on top of the page title.

Pageheader was renamed to header, so all that functionality should still be there.

pbteja1998:
> > @normdoow @hcorta `PageHeader` is something different and I am using it. Any reason why it got removed here? CleanShot 2021-12-18 at 07 59 35@2x
> > Page header previously used to be on top of the page title.
> 
> Pageheader was renamed to header, so all that functionality should still be there.

Oh... I thought `header` was about replacing breadcrumbs.

hcorta:
@pbteja1998 are you referring to the "pageHeader" prop, or the component "PageHeader" within the project?

The pageHeader prop and its use has not been changed at all.

The PageHeader component was not exposed previously, but now it is through the header prop, that allows you to pass a component and resolve any custom case, instead of having to do internal hacks.

Hope that helps☺️

PR #189, opened 2022-01-12
added the block for page links

normdoow:
Only allows you to link to pages that are in the same workspace.
Previously we had supported page links though the page mention block and text block page links.
This PR actually supports the Notion Page Link block as well.

Example of page link block here: 52353862df0f48ba85648db7d0acd1dd

image

### DIFF
--- a/packages/react-notion-x/src/block.tsx
+++ b/packages/react-notion-x/src/block.tsx
@@ -815,6 +815,23 @@ export const Block: React.FC<BlockProps> = (props) => {
     case 'transclusion_reference':
       return <SyncPointerBlock block={block} level={level + 1} {...props} />
 
+    case 'alias':
+      let blockPointerId = block?.format?.alias_pointer?.id
+      const linkedBlock = recordMap.block[blockPointerId]?.value
+      if (!linkedBlock) {
+        console.log('"p" missing block', blockPointerId)
+        return null
+      }
+
+      return (
+        <components.pageLink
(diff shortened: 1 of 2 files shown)


packages/react-notion-x/src/block.tsx
```diff
@@ -815,6 +815,23 @@ export const Block: React.FC<BlockProps> = (props) => {
     case 'transclusion_reference':
       return <SyncPointerBlock block={block} level={level + 1} {...props} />
 
+    case 'alias':
+      let blockPointerId = block?.format?.alias_pointer?.id
```
transitive-bullshit, line 819: try to use `const` if a variable doesn't change; you may need to surround the case block in braces (this is a small nit; just browsing 😄 )

transitive-bullshit (approved):
Looks great dude 💪

packages/react-notion-x/src/block.tsx
normdoow: I'll update this! Thanks for taking a look!

normdoow:
Live on v4.13.2!

Mapaor:
Great! Although it does not work for mentioned page blocks. So if inside Notion I mention a heading of a page, then that block in react-notion-x gets mentioned as a separate page and with no styles applied. Is there a way to fix this?

Notion:
block-mention
page-block-notion

react-notion-x:
notion-x-block-mention
block-page

Issue #166, opened 2021-11-24
Support for simple tables


MartinXPN:
In one of their latest releases, the Notion team released a feature that adds support for simple tables. This blogpost describes the feature in more details: https://www.notion.so/guides/simple-tables-vs-databases

Currently, when loading a page containing a simple table, `react-notion-x` does not render anything and the space for the table is completely empty. 

Are there plans to support the newly added feature?

Thanks for your work and the great library!

thanhlmm:
Yeah, I'm also waiting for simple table supported

normdoow:
If someone wanted to add a PR for this, that would be great!

normdoow:
Live for v4.14.0

MartinXPN:
Seems like all the styles are completely off after upgrading.
All the code blocks, headings, etc used to render completely fine, but all of them lost their styling after the upgrade. 
Am I missing something?

normdoow:
> Seems like all the styles are completely off after upgrading. All the code blocks, headings, etc used to render completely fine, but all of them lost their styling after the upgrade. Am I missing something?
> 
> The last version I was using was v4.11.0

Hmm maybe. Make sure you are updating all the packages. notion-types, notion-utils, notion-client, and react-notion-x to v4.14.1 (.1 has a fix for default tables).

MartinXPN:
Yeah, you were right, `yarn remove`-ing the package and then `yarn add`-ing it resolved the issue.
Thanks!

PR #191, opened 2022-01-15
Add support for simple table

enpitsuLin:
inspired by #166 

add block type and also add some simple stylesheet

simple-table

***
also noticed default css variable is not variety enough just like `fg-color-1(~5)`

### DIFF
--- a/packages/react-notion-x/src/block.tsx
+++ b/packages/react-notion-x/src/block.tsx
@@ -832,6 +832,32 @@ export const Block: React.FC<BlockProps> = (props) => {
         </components.pageLink>
       )
 
+    case 'table':
+      return (
+        <table className='notion-simple-table'>
+          <tbody>{children}</tbody>
+        </table>
+      )
+    case 'table_row':
+      const tableBlock = recordMap.block[block.parent_id]
+        .value as types.TableBlock
+      const order = tableBlock.format.table_block_column_order
+      const formatMap = tableBlock.format.table_block_column_format
+
+      return (
+        <tr className='notion-simple-table-row'>
+          {order.map((column) => {
+            return (
+              <td key={column} style={{ width: formatMap[column].width }}>
+                <div className='notion-simple-table-cell'>
+                  {block.properties[column]}
+                </div>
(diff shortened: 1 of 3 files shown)


normdoow:
Hey @enpitsuLin, thanks for adding this. Awesome! I will take a look in the next couple of days.

imsukmin:
When did you deploy it? I need this function.

normdoow:
> When did you deploy it? I need this function.

This hasn't been deployed yet. But Hopefully tomorrow if everything is good.

packages/react-notion-x/src/block.tsx
```diff
@@ -832,6 +832,37 @@ export const Block: React.FC<BlockProps> = (props) => {
         </components.pageLink>
       )
 
+    case 'table':
+      return (
+        <table className='notion-simple-table'>
```
normdoow, line 837: Can you add the notion blockId as a class? Like `className={cs(blockId)}`. You'll see most of the other blocks also have this. This is so people can target specific blocks and customize it with css.

packages/react-notion-x/src/block.tsx
```diff
@@ -832,6 +832,37 @@ export const Block: React.FC<BlockProps> = (props) => {
...
+        </table>
+      )
+    case 'table_row':
+      const tableBlock = recordMap.block[block.parent_id]
+        .value as types.TableBlock
+      const order = tableBlock.format.table_block_column_order
+      const formatMap = tableBlock.format.table_block_column_format
+
+      return (
+        <tr className='notion-simple-table-row'>
```
normdoow, line 848: Can you add the notion block id to the rows as well? `className={cs(blockId)}`.

packages/react-notion-x/src/block.tsx
enpitsuLin: got it.just finished

normdoow:
Live for v4.14.0! Thanks @enpitsuLin.
One thing I did see was missing that would be nice to add is styling based on the table row header and column header setting in Notion.

Issue #62, opened 2021-05-08
Prop `className` did not match


whitedelay:
Hello, thank you for your great project.

I have an issue.
I rendered my notion page using `react-notion-x` in my `next.js` project, and this error sometimes occurred.
```
Warning: Prop "className" did not match. Server: " language-bash" Client: "language-bash"
```

I found that the class name of code block has space at the front.
```
<pre class="notion-code language-bash">
  <code class=" language-bash">
```

Only `code` block has this issue. 
Do you know how to solve this problem?

MichaelDevC:
I have the same error, still don't know how to solve it...

remorses:
Reading other prismjs examples like this [one](https://github.com/satya164/react-simple-code-editor/blob/e5be544449a7ed0387a14236fff1c41bc6f0798f/src/index.js#L586) i think we can just skip adding the `landuage-{language}` classname, it looks like prism doesn't need that when using the `highlight` function

transitive-bullshit:
The `Code` component has been updated to fix this issue in v6.

PR #194, opened 2022-01-19
fix: Prop className did not match of Code component

futantan:
This PR fix the issue: #62 

[the \<pre\> will automatically get the language-xxxx class (if it doesn’t already have it) and will be styled as a code block.](https://prismjs.com/#:~:text=the%20%3Cpre%3E%20will%20automatically%20get%20the%20language%2Dxxxx%20class%20(if%20it%20doesn%E2%80%99t%20already%20have%20it)%20and%20will%20be%20styled%20as%20a%20code%20block.)

### DIFF
--- a/packages/react-notion-x/src/components/code.tsx
+++ b/packages/react-notion-x/src/components/code.tsx
@@ -11,7 +11,7 @@ export const Code: React.FC<{ code: string; language: string }> = ({
   const prismLanguage = languages[languageL] || languages.javascript
 
   return (
-    <pre className='notion-code'>
+    <pre className={`notion-code language-${languageL}`}>
       <code
         className={`language-${languageL}`}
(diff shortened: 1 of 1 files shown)


transitive-bullshit:
This is awesome @futantan; thanks so much.

kujyp:
@transitive-bullshit 
This PR code was removed by refactoring commit bbb899da21f95957069ad6639cbf88063004cd19

Before:
https://github.com/NotionX/react-notion-x/blob/100dc9191390111b52ac27b9ea31f17451100fff/packages/react-notion-x/src/components/code.tsx#L14

After:
https://github.com/NotionX/react-notion-x/blob/bbb899da21f95957069ad6639cbf88063004cd19/packages/react-notion-x/src/third-party/code.tsx#L76

PR #198, opened 2022-01-19
fix: notion simple table when there are empty cells

pbteja1998:
Simple Table is not working when there are empty cells in the table.


normdoow (approved):
Thanks for this @pbteja1998!

normdoow:
v4.14.2!

PR #204, opened 2022-01-27
add non-text support for getpageProperty 

enpitsuLin:
support which marked with TODO 😁

Actually, the param `block` is not much useful (maybe `pageId`  which was `block.parent_id` would better), reserve it for non-breaking changes

***
also done an unit test later, but few of wrong type definitions which need be fix prevent me from using my test cases


normdoow (approved):
This is a nice add. Thanks for this!

PR #206, opened 2022-01-27
added support for toggle header blocks

normdoow:
PageId: 5995506f2c564d81956aa38711e12337

Header blocks now handle the toggle format.
image


normdoow:
Live on v4.15.0

PR #215, opened 2022-02-16
Optimizing images

normdoow:
PageId: 3492bd6dbaf44fe7a5cac62c5d402f06

Using the block width to grab the correct image size and optimize images.

image

### DIFF
--- a/packages/react-notion-x/src/components/asset.tsx
+++ b/packages/react-notion-x/src/components/asset.tsx
@@ -223,6 +223,7 @@ export const Asset: React.FC<{
         alt={alt}
         zoomable={false}
         height={style.height as number}
+        width={block?.format?.block_width}
         style={assetStyle}
       />
     )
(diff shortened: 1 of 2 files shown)


transitive-bullshit:
@normdoow this looks like a solid change, though I'm not in a spot to test it atm.

transitive-bullshit:
@normdoow is this intended for unsplash images only or notion-hosted images?

I don't think we can always add the `&w=` parameter because it may break some use cases.

normdoow:
> 

yeah, you're probably right. The idea for this came from a Notion employee actually. they do this for some of their images though I'm not sure that it is supported for all.

_On the subject of Pagespeed's image guidance, Notion gets one out of two correct - when viewing these images from a Notion doc, we request & serve images at their displayed size using a query parameter, but we don't currently transcode the image to a more efficient format based on the browser's image support._

PR #222, opened 2022-03-04
fixed date being one off for properties and formulas

normdoow:
Notion's data is based off of UTC but before the Date would transform to the local time.
Made a change to make sure we are saving in UTC time.

Page to test: 43fb2254b7e3412ca461566f7a83918e

Before: was the 22nd even though it should be 23rd.
image

Now:
image


normdoow:
live in v4.16.2

PR #223, opened 2022-03-07
Widen horizontal padding on small screens

kidonng:
On a iPhone 12 Pro screen, horizontal padding is calculated as 7.8px which looks awful for such content-rich website:

image

On super.so the minimum padding is 24px. That's a lot so 16px would do it.


transitive-bullshit (approved):
Great change; thanks for the explanation @kidonng

transitive-bullshit:
Published in `4.17.0`

PR #224, opened 2022-03-16
Show the checkbox text on collection card

uduakabaci:
The library only shows a checkbox on a collection card without the checkbox's name. 
**This is what it looks like currently** 
check-list-w-o-text

**And this what it looks like on notion**
check-list-w-text

I've been able to implement this by adding the checkbox name [from the schema] to the collection cards.  Here is a link to the notion page  https://talented-work-a0b.notion.site/Checklist-128b92cbe975470c8a8594ad232cf3da


transitive-bullshit (approved):
This looks great @uduakabaci. Thanks for the help 🙏

transitive-bullshit:
Published in `4.17.0`

PR #225, opened 2022-03-19
feat: add page content only feature

ntcho:
First, thank you for your work in such a solid library.

## Why is this change needed?

I liked the theme which automatically comes with `NotionRenderer`, but I needed a way to get contents of a page as a raw HTML to customize it furtuer, without the extra components like title, footer, cover, etc.

## How is this change used?

Previously, the following code will render with the default theme:

<table>
<tr><td>Code</td><td>Preview</td></tr>
<tr>
<td>

```jsx
<NotionRenderer
  recordMap={recordMap}
  rootPageId={site.rootNotionPageId}
  ...
/>
```

</td>
<td>

image

</td>
</tr>
</table>

With the change, you can add `pageContentOnly={true}` as a property and the `NotionRenderer` component will render in raw HTML:

<table>
<tr><td>Code</td><td>Preview</td></tr>
<tr>
<td>

```jsx
<NotionRenderer
  recordMap={recordMap}
  rootPageId={site.rootNotionPageId}
  pageContentOnly={true}
  ...
/>
```

</td>
<td>

image

</td>
</tr>
</table>

## Notion Page used

Page used is my page [here](https://www.notion.so/ntcho/About-f19c6d673e474fc2894862dee9d201fe), duplicated from Travis Fischer's original notion page.

### DIFF
--- a/packages/react-notion-x/src/block.tsx
+++ b/packages/react-notion-x/src/block.tsx
@@ -40,6 +40,7 @@ interface BlockProps {
   pageFooter?: React.ReactNode
   pageAside?: React.ReactNode
   pageCover?: React.ReactNode
+  pageContentOnly?: boolean
 
   hideBlockId?: boolean
   disableHeader?: boolean
@@ -76,6 +77,7 @@ export const Block: React.FC<BlockProps> = (props) => {
     pageFooter,
     pageAside,
     pageCover,
+    pageContentOnly,
     hideBlockId,
     disableHeader
   } = props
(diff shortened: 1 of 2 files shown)


pbteja1998:
Isn't this same as setting `fullPage={false}`?

ntcho:
Oh I see 😓
Didn't know `fullPage` props meant full width page as it means in Notion.

Will close this PR. Thanks!

pbteja1998:
Yeah, full width pages will work automatically based pm whether it is full width on Notion or not. You don't need to manually control the full width setting in your code.

transitive-bullshit:
@ntcho really appreciate your work but as @pbteja1998 says this should already be supported.

I agree this is a really important use case, however. Maybe we could rename `fullPage` or add it to the docs so it's more clear what it does?

ntcho:
Thanks! I've confirmed what the `fullPage` props is what I was looking for.

I think it will be more clear to use `pageChildrenOnly` as the property alias or briefly explain what each prop does in the prop type interface.

Issue #226, opened 2022-03-19
Inline galleries do not work


janezturk:
Thank you very much for awesome library!

I think Notion changed something regarding inline galleries. 
Looks like NotionClient is not returning inline gallery as a part of the page.

Here is my example page:
https://tall-flock-373.notion.site/Problem-with-gallery-page-8a4e3d43f7f54323a2d4aca5647a6b5a

ntcho:
It seems the recent update of inline database styles are causing this problem. 🙁

Dineshs91:
I have noticed the same.

ntcho:
`CollectionViewGallery`, `CollectionViewList`, `CollectionViewTable` are affected by this change.

The problem seems to be originated from the modification of the props being passed to these objects:

<table>
<tr><td>Previous props</td><td>Current props</td></tr>
<tr>
<td>

```js
{
  collection: {
    id: 'ba828d5c-47bd-4c26-...',
    ...
  },
  collectionView: {
    id: '0981a186-f238-44ac-...',
    ...
  },
  collectionData: {
    type: 'results',
    collection_group_results: {
      blockIds: [
        'c6bcd16d-ff46-4b7c-...',
      ],
    },
    hasMore: false
  },
  padding: 96,
  width: 1024
}
```

</td>
<td>

```js
{
  collection: {
    id: 'ba828d5c-47bd-4c26-...',
    ...
  },
  collectionView: {
    id: '0981a186-f238-44ac-...',
    ...
  },
  collectionData: {
    type: 'results',
    blockIds: [
      'c6bcd16d-ff46-4b7c-...',
    ],
    hasMore: false
  },
  padding: 96,
  width: 1024
}
```

</td>
</tr>
</table>

### Quick fix

Replace all `collectionData['collection_group_results'].blockId` to `collectionData.blockId` in JS files located in `/node_modules/react-notion-x/build/ejs` and `/node_modules/react-notion-x/build/esm` directiory

transitive-bullshit:
This should be fixed in `v6.0.7`

PR #227, opened 2022-03-20
fix: inline gallery undefined error

ntcho:
This pull request fixes NotionX/react-notion-x#226.

## Current problem

`CollectionViewGallery`, `CollectionViewList`, `CollectionViewTable` are affected by external source (can't confirm, but possibly related to Notion's latest change in API with the collection view design update).

The problem seems to be originated from the modification of the props being passed to these objects:

<table>
<tr><td>Previous props</td><td>Current props</td></tr>
<tr>
<td>

```js
{
  collection: {
    id: 'ba828d5c-47bd-4c26-...',
    ...
  },
  collectionView: {
    id: '0981a186-f238-44ac-...',
    ...
  },
  collectionData: {
    type: 'results',
    collection_group_results: {
      blockIds: [
        'c6bcd16d-ff46-4b7c-...',
      ],
    },
    hasMore: false
  },
  padding: 96,
  width: 1024
}
```

</td>
<td>

```js
{
  collection: {
    id: 'ba828d5c-47bd-4c26-...',
    ...
  },
  collectionView: {
    id: '0981a186-f238-44ac-...',
    ...
  },
  collectionData: {
    type: 'results',
    blockIds: [
      'c6bcd16d-ff46-4b7c-...',
    ],
    hasMore: false
  },
  padding: 96,
  width: 1024
}
```

</td>
</tr>
</table>

## Changes made

Replaced all occurrences referencing `collectionData['collection_group_results'].blockIds` into `collectionData.blockIds`.

> I couldn't build the source and test this change myself, but applying this modification directly to the build files fixed this issue. Please test this change before merging it.

## Notion page for testing

https://tall-flock-373.notion.site/Problem-with-gallery-page-8a4e3d43f7f54323a2d4aca5647a6b5a (from #226)

### DIFF
--- a/packages/react-notion-x/src/components/collection-view-gallery.tsx
+++ b/packages/react-notion-x/src/components/collection-view-gallery.tsx
@@ -30,7 +30,7 @@ export const CollectionViewGallery: React.FC<CollectionViewProps> = ({
     <Gallery
       collectionView={collectionView}
       collection={collection}
-      blockIds={collectionData['collection_group_results'].blockIds}
+      blockIds={collectionData.blockIds}
     />
   )
 }
--- a/packages/react-notion-x/src/components/collection-view-list.tsx
+++ b/packages/react-notion-x/src/components/collection-view-list.tsx
@@ -28,7 +28,7 @@ export const CollectionViewList: React.FC<CollectionViewProps> = ({
 
   return (
     <List
-      blockIds={collectionData['collection_group_results'].blockIds}
+      blockIds={collectionData.blockIds}
       collection={collection}
       collectionView={collectionView}
     />
--- a/packages/react-notion-x/src/components/collection-view-table.tsx
+++ b/packages/react-notion-x/src/components/collection-view-table.tsx
@@ -43,7 +43,7 @@ export const CollectionViewTable: React.FC<CollectionViewProps> = ({
 
   return (
     <Table
-      blockIds={collectionData['collection_group_results'].blockIds}
+      blockIds={collectionData.blockIds}
       collection={collection}
       collectionView={collectionView}
       padding={padding}


transitive-bullshit:
This looks really great @ntcho.

Any idea if this change affects our API calls https://github.com/NotionX/react-notion-x/blob/master/packages/notion-client/src/notion-api.ts#L302-L314?

ntcho:
Closed by accident 😅

I haven't read the `notion-client` package in depth yet, but I will add a comment here if I find any change that affects the API. So far, the pages rendered as expected for me.

transitive-bullshit:
Tested locally and I can repro the issue on your page, but if I duplicate the page into my workspace, the issue goes away (https://www.notion.so/Inline-Gallery-9cb9716c93164c6c8b4cd0bac3879aeb).

I'm also not seeing gallery issues on previous instances, so I'm guessing this may only affect gallery instances under certain conditions.

So I think we should update this PR to be a bit more conservative and check both ways of accessing the data — so it works in all cases.

@ntcho Thoughts?

ntcho:
I see. I agree that this PR should be tested and updated thoroughly.

@transitive-bullshit Also, it would be helpful if you can share me your workflow on how you build and test the packages. I'm only testing the changes via directly editing files in `build` file because I don't know how to build it myself 😭

transitive-bullshit:
@ntcho here you go https://github.com/NotionX/react-notion-x/blob/master/contributing.md 💪

ntcho:
Thanks! I'll look into it and add further commits or comments if I find anything new.

normdoow:
I'm seeing this issue on new databases that are created but not old ones.
So I think we could have an optional check so we can do either.
How does that sound?

transitive-bullshit:
Thanks for the suggested fix @ntcho. I ended up adding this in 4c974dc and changed things to make sure it works with both ways of accessing the data.

Released as `v6.0.7`.

PR #229, opened 2022-03-21
General maintenance update

transitive-bullshit:
- lots of misc bug fixes
- update some core deps (typescript, next.js, etc)
- fix custom image support focusing on `next/image` use case 🔥 
- normalize how preview images are loaded and supported for both custom and non-custom images
- re-enable and fix `medium-zoom`
- update the next.js example to use a custom image with preview images enabled via [lqip-modern](https://github.com/transitive-bullshit/lqip-modern)

@normdoow @pbteja1998 just a head's up that this is a larger update

### DIFF
--- a/example/lib/preview-images.ts
+++ b/example/lib/preview-images.ts
@@ -1,0 +1,74 @@
+import got from 'got'
+import lqip from 'lqip-modern'
+import pMap from 'p-map'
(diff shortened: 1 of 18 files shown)


normdoow:
Awesome! Thanks @transitive-bullshit.

PR #245, opened 2022-03-28
feat: support optional custom embed component

ciiiii:
Implementation about issue https://github.com/NotionX/react-notion-x/issues/240.


transitive-bullshit (approved):

transitive-bullshit:
This looks great; thanks @ciiiii!

The only thing I could see us wanting as a follow-up would be to expose `AssetWrapper` as an export so custom embeds could fall-back to the default functionality if they want.

PR #248, opened 2022-03-31
fix response 401 due to auth token not passed as bearer token

wenwkich:
This update should fix the issue that passing `authToken` but still getting 401 response

It should be passed as `Authorization` bearer token according to this notion page:
https://developers.notion.com/docs/authorization

### DIFF
--- a/packages/notion-client/src/notion-api.ts
+++ b/packages/notion-client/src/notion-api.ts
@@ -543,7 +543,7 @@ export class NotionAPI {
     }
 
     if (this._authToken) {
-      headers.cookie = `token_v2=${this._authToken}`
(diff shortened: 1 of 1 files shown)


AakashMallik:
@transitive-bullshit Could you please review this whenever you are available? This will fix a major bug with private pages.

transitive-bullshit:
I don't believe this is how the notion private API is intended to be authorized (which is what `notion-client` uses). If something changed there, then it will take some time for me to test.

If you're talking about the official, public notion API, then this PR doesn't apply. Check out https://github.com/NotionX/react-notion-x/pull/256 for my WIP on supporting it.

Note, however, that using the official, public notion API is not well supported since a lot of blocks don't contain enough info to display properly.

PR #261, opened 2022-04-07
match simple table column widths with Notion

pbteja1998:
#### Description

In Notion, Cells of Simple Table has `min-width` and `max-width` instead of `width`. 

So, changed the styles as per Notion styles.

#### Notion Test Page ID
1e739b87549e406a9461d2a7e4ff7845

### DIFF
--- a/packages/react-notion-x/src/block.tsx
+++ b/packages/react-notion-x/src/block.tsx
@@ -794,7 +794,8 @@ export const Block: React.FC<BlockProps> = (props) => {
                 key={column}
                 className={color ? `notion-${color}` : ''}
                 style={{
-                  width: formatMap?.[column]?.width || 120
+                  minWidth: 120,
+                  maxWidth: 240
                 }}
               >
                 <div className='notion-simple-table-cell'>


packages/react-notion-x/src/block.tsx
```diff
@@ -794,7 +794,8 @@ export const Block: React.FC<BlockProps> = (props) => {
                 key={column}
                 className={color ? `notion-${color}` : ''}
                 style={{
-                  width: formatMap?.[column]?.width || 120
```
transitive-bullshit, line 797: Can `formatMap[column].width` exist?

transitive-bullshit:
This looks great @pbteja1998. Thanks! 🙏 

Just one small question before merging.

packages/react-notion-x/src/block.tsx
pbteja1998: I'm not exactly sure. I know that columns cannot be resized manually. 

Will do one thing. Will keep the width property too, but instead of having the fallback width as `120`, I will keep it as `auto`.

PR #265, opened 2022-04-11
Fixed databases for new Notion views

normdoow:
#### Description
Hot fix for the new databases. Continues to work for old database versions as well.
Hopefully this is a pretty good step in the right direction.

The new filters do not work yet. Though I did add some TODOs with some code that seemed to be close to getting them there.

DBs with Multiple views are working but there is a shortcoming. All the view have to be linked to the same DB. If one of the views is linked to another db, it won't work currenlty.

Screenshots.
image
image
image

#### Notion Test Page ID
fcb43d800f2b4a47bca1c1b51281e598
This page includes sub pages of multiple new Notion databases, inline and full DBs.


normdoow:
@transitive-bullshit my package-lock.json has a lot of changes. Not sure why it does so i didnt' push it. But that could be why the build-full is failing.

transitive-bullshit:
@normdoow there shouldn't be a package-lock.json if you're using yarn v1. I'd rather not switch package managers at this time.

normdoow:
> 

Oh that's right. I think I accidentally did a npm install because I was moving from a different project. Forgot it was yarn initially.

packages/notion-utils/src/get-block-collection-pointer-id.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  CollectionViewBlock,
+  CollectionViewPageBlock,
+  ExtendedRecordMap
+} from 'notion-types'
+
+export function getBlockCollectionPointerId(
```
transitive-bullshit, line 7: i think this logic could be inside of `getBlockCollectionId` instead of exporting a new method. thoughts?

transitive-bullshit (approved):

packages/notion-utils/src/get-block-collection-pointer-id.ts
normdoow: So just have an if statement in there based on if the recordmap is passed in? Either way seems fine to me.

normdoow: Yep, that looks good!

PR #278, opened 2022-04-26
Remove styles inconsistency

Commandtechno:
This style breaks dividers, and only applies when `hideBlockId` is enabled

#### Description

This padding adds some weird borders around the sides of a divider

image

image
 
It also only applies when `hideBlockId` is enabled since it doesn't match `notion-block-{ID}`

image

image

#### Notion Test Page ID

`ID` 0be6efce9daf42688f65c76b89f8eb27
`URL` https://react-notion-x-demo.transitivebullsh.it/0be6efce9daf42688f65c76b89f8eb27

To reproduce, remove the ID from the divider's `notion-block` class name which is what `hideBlockId` would set it to

### DIFF
--- a/packages/react-notion-x/src/styles.css
+++ b/packages/react-notion-x/src/styles.css
@@ -834,10 +834,6 @@ svg.notion-page-icon {
   flex-direction: column;
 }
 
-.notion-block {
-  padding: 3px 2px;
-}
-
 .notion .notion-code {
   font-size: 85%;
 }


Commandtechno:
Adding on to this, is there any reason for the `notion-block` class name since it isn't used anywhere else?

transitive-bullshit:
> Adding on to this, is there any reason for the notion-block class name since it isn't used anywhere else, and doesn't affect anything without the hideBlockId option currently?

It's probably not necessary. `react-notion-x` has gone through a lot of refactors and reverse-engineering styles from notion (which has also changed over time) is not an easy thing to do cleanly.

This looks great — thanks!

transitive-bullshit:
> Adding on to this, is there any reason for the notion-block class name since it isn't used anywhere else, and doesn't affect anything without the hideBlockId option currently?

It's probably not necessary. `react-notion-x` has gone through a lot of refactors and reverse-engineering styles from notion (which has also changed over time) is not an easy thing to do cleanly.

This looks great — thanks!

transitive-bullshit:
Released as `6.12.6`

PR #281, opened 2022-04-30
Add 'displayMode' in katexSettings

Thooooor:
#### Notion Test Page ID
PINN-df5c77e6831d4c2d9f9fa81f18bccdf0


jacksongoode:
@Thooooor I tried this out as a patch for the library and it didn't seems to resolve cases where `\begin{align}...\end{align}` appeared in the math blocks. Hmm, any thoughts?

jacksongoode:
It might be useful to also add an example of an equation that would require the `displayMode=True` within the preview demo test?

jacksongoode:
@transitive-bullshit This one might be an easy fix for many broken equation blocks.

PR #282, opened 2022-05-01
feat(add support for private Notion page assets)

jmuchovej:
#### Description
Traverses a page's blocks (for `collection_view` support) to generate
Signed URLs for images. Also expands `defaultMapImageUrl` to support
lookups for UGC (user-generated content) on Page Covers and Page Icons.
(Which will follow a form similar to
 `https://notion.so/<UUID>/<filename>.<ext>`. Since the OG
 implementation maps these assets to `block.id`, which refers to the
 Page, only one of Cover / Icon were allowed to be UGC.)

#### Notion Test Page ID

I used the shipped examples in `examples/full`, so `067dd719a912471ea9a3ac10710e7fdf`.

This is intended to work on private pages and didn't appear to alter public pages.

**Note:** It _seems_ that image loading is slower from AWS now. **But**, I believe this is due to one of the GIFs on my test page being ~30MB. 😅

### DIFF
--- a/packages/notion-client/src/notion-api.ts
+++ b/packages/notion-client/src/notion-api.ts
@@ -6,7 +6,8 @@ import {
   parsePageId,
   getPageContentBlockIds,
   uuidToId,
-  getBlockCollectionId
+  getBlockCollectionId,
+  isUrl
 } from 'notion-utils'
 import * as notion from 'notion-types'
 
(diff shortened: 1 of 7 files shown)


jmuchovej:
@transitive-bullshit Since you refactored the `map-*-url` into their own files – should I do so here, too?

jmuchovej:
@transitive-bullshit A friendly ping about this. 🙂 Anything I can do to speed things along?

PR #309, opened 2022-06-24
feat(docs): add explanation for notion-client

bytrangle:
#### Description

I think the current documentation for `notion-client` doesn't do enough justice to how powerful it is.

I explain the shape of the response that `getPage` method is supposed to return and give example response.

#### Notion Test Page ID
cc368b47772a4a1aa36e1f52c507d20d


transitive-bullshit (approved):

transitive-bullshit:
Thanks @bytrangle 🙏

PR #312, opened 2022-06-27
Use fetch instead of got in `notion-client`

### LINKED ISSUE #160: Alternative to notion-client format?
Great library. Loving it so far.

My use case: Wanting to render page server side inside of Cloudflare Workers.  Rendering is working (thanks to use of flareact.com)

Only piece not working is the notion-client.  I can't use that library within Cloudflare Workers due to dependance of 'got' library which doesnt run in Cloudflare.

I noticed you had the fetch() implemented originally but is commented out.  I'm wondering if I fork notion-client and use fetch (which does work) instead of got if that would work for me?

Or, I can use the Notion API Worker from SplitBee, but it doesn't look like their format is supported by react-notion-x. Is that correct? Is there a way to convert their recordMap format to what is needed for rendering with x?

remorses:
#### Description

Using fetch from undici or the native fetch implementation, this makes it possible to use notion-client in `cloudflare workers` or other environments other than node

Given that this is a breaking change (renamed `gotOptions` to `fetchOptions`) it could need a major version bump, tell me if you want me to change the version of the packages

Fix #160

### DIFF
--- a/packages/notion-client/package.json
+++ b/packages/notion-client/package.json
@@ -22,6 +22,7 @@
     "test": "ava"
   },
   "dependencies": {
+    "node-fetch": "^3.2.6",
     "notion-types": "^6.12.6",
     "notion-utils": "^6.12.9",
     "p-map": "^5.3.0",
--- a/packages/notion-client/package.json
+++ b/packages/notion-client/package.json
@@ -22,10 +22,10 @@
     "test": "ava"
   },
   "dependencies": {
-    "got": "^11.8.1",
     "notion-types": "^6.12.6",
     "notion-utils": "^6.12.9",
-    "p-map": "^5.3.0"
+    "p-map": "^5.3.0",
+    "undici": "^5.5.1"
   },
   "ava": {
     "snapshotDir": ".snapshots",
--- a/packages/notion-client/src/notion-api.ts
+++ b/packages/notion-client/src/notion-api.ts
@@ -1,5 +1,5 @@
 // import { promises as fs } from 'fs'
-import got, { OptionsOfJSONResponseBody } from 'got'
+
 import pMap from 'p-map'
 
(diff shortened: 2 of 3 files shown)


pbteja1998:
Yes, even I use this cloudflare workers and i had to add a patch everytime to replace got with fetch.

transitive-bullshit:
Really great PR @remorses. Much appreciated 🙏 

My main hesitancy here is that the notion API can be pretty flaky at times, and `got` includes some really nice, non-trivial retry logic by default. I'm worried that existing use cases would become less stable with this change.

Maybe we could allow for the user to pass in their own `fetch` implementation to the client constructor?

I'd still like `got` to be usable without any additional work on the user's part. Maybe we could move it to an [optional peer dependency](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta) and try conditionally importing it if it's available? If not, then we fallback to the `fetch` logic in this PR?

remorses:
What if we move the retry logic inside the fetch method? The user could then pass something like `fetchOptions: {retry: 3}`

transitive-bullshit:
I've finally had a chance to revisit this while trying to get `notion-client` to work with the new Vercel edge runtime.

The main issue I'm running into is that including `node-fetch` anywhere in the bundle (even conditionally imported) means that it doesn't work on non-node.js runtimes such as Vercel's edge runtime. I haven't tried with CF workers, but if it works there, then I'd guess the difference is Vercel's eager bundling optimizations.

Either way, I'd love to support this use case for `notion-client`, but am not sure of how to proceed aside from having the user pass in a `fetch` implementation which is kind of poor DX.

transitive-bullshit:
Resolved in https://github.com/NotionX/react-notion-x/pull/394

PR #319, opened 2022-07-02
fix TypeError in addSignedUrls

remorses:
#### Description

In some cases a content page id is not in the block object, this can cause a `TypeError: Cannot read property 'value' of undefined`


transitive-bullshit:
Thanks! 🙏

PR #320, opened 2022-07-03
Fix youtube embed starting at specific timestamp

### LINKED ISSUE #321: Youtube embedded video with specific timestamps doesn't open on said timestamp when played
#### Youtube embedded video with specific timestamps doesn't open on said timestamp when played

I've noticed that embedding a video to notion page with a specific timestamp is properly supported by notion itself, but when said page is rendered using react-notion-x renderer, the video simply starts from the beginning.

#### Notion Test Page ID -> 769e7a8e652a42518a9780e804ae97cf

martin-lacko-rebood:
#### Youtube embedded video with specific timestamps doesn't open on said timestamp when played

I've noticed that embedding a video to notion page with a specific timestamp is properly supported by notion itself, but when said page is rendered using react-notion-x renderer, the video simply starts from the beginning.

I believe this is not expected behaviour so I had a look. It appears the component used to render the youtube video (`LiteYouTubeEmbed`) itself does support `params`, however those params were actually never passed to said component. I added simple parse function for the URL search params that returns them as a `Record<string, string>` and pass those params down to the component which solves the issue. I'm calling the function itself `getUrlParams` not `getYoutubeUrlParams` because it has no relation to youtube URLs themselves, it simply returns `params` from whatever url is provided thus can be used elsewhere as well.

#### Notion Test Page ID -> 769e7a8e652a42518a9780e804ae97cf

Fixes #321


transitive-bullshit:
Thanks @martin-lacko-rebood 🙏

PR #328, opened 2022-07-12
Fixed new uncategorized relation that can be used in Gallery Views

normdoow:
#### Description
Notion has another data field for default uncategorized groups that we don't support currently.
image

#### Notion Test Page ID
ae22f7841d3c45668a08f56d1bd26a98

In this example these gallery items weren't being shown because the blockIds weren't being found correctly so nothing was being shown.
image


transitive-bullshit:
🚀

PR #333, opened 2022-07-26
FEAT: Added support for the newest version of database filters

normdoow:
#### Description

Notion changes how their databases work a while back and updated how users can filter them. This update adds support for that newest update.

Now notion-react-x can show databases even if they have filters on them. This should work for all types of views.

#### Notion Test Page ID

multiple different text examples including older version of databases that still work.

84696b02b91d45a4a879ff572c23cc93
3271d83e87a14a42b5a53fd16c4339a7?v=ef3aeeb35e3045dc9b04abba22c10648
6f6138a8d9ab4e2b85a0f664d035b9a3?v=ca9b476a978d430a81ceac957b28dee7
a0cb6346eaba4a4f930a4c4206942b1e

image
image

### DIFF
--- a/packages/notion-client/src/notion-api.ts
+++ b/packages/notion-client/src/notion-api.ts
@@ -332,6 +332,18 @@ export class NotionAPI {
       collectionView?.format?.board_columns_by ||
       collectionView?.format?.collection_group_by
 
+    let filters = []
+    if (collectionView.format?.property_filters) {
+      filters = collectionView.format?.property_filters.map((filterObj) => {
+        // console.log('map filter', filterObj)
+        //get the inner filter
+        return {
+          filter: filterObj.filter.filter,
+          property: filterObj.filter.property
+        }
+      })
+    }
+
     let loader: any = {
       type: 'reducer',
       reducers: {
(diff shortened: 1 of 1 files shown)


packages/notion-client/src/notion-api.ts
```diff
@@ -332,6 +332,18 @@ export class NotionAPI {
       collectionView?.format?.board_columns_by ||
       collectionView?.format?.collection_group_by
 
+    let filters = []
+    if (collectionView.format?.property_filters) {
+      filters = collectionView.format?.property_filters.map((filterObj) => {
+        // console.log('map filter', filterObj)
+        //get the inner filter
+        return {
+          filter: filterObj.filter.filter,
```
transitive-bullshit, line 341: Let's make sure to use `?.` everywhere here just to make sure the code doesn't crash. I generally don't trust notion's data model to be consistent across future updates.

transitive-bullshit (changes requested):
Looks great. Just one small request.

packages/notion-client/src/notion-api.ts
normdoow: I don't trust their model either. 😁
Good thinking. updated this.

transitive-bullshit (approved):

transitive-bullshit:
Thanks Noah! 🙏

PR #342, opened 2022-08-04
Use <p> tag for text blocks

### LINKED ISSUE #341: <p> Tags over <div> for text?
#### Description
Currently, the text blocks are rendered as `<div>` tags. Wouldn't it be preferable to render those with `<p>` tags instead?

MartinXPN:
#### Description
Fixes #341

With this PR, the text blocks are rendered as `<p>` tags in the generated HTML instead of a `<div>`.

### DIFF
--- a/packages/react-notion-x/src/block.tsx
+++ b/packages/react-notion-x/src/block.tsx
@@ -405,7 +405,7 @@ export const Block: React.FC<BlockProps> = (props) => {
       const blockColor = block.format?.block_color
 
       return (
-        <div
+        <p
           className={cs(
             'notion-text',
             blockColor && `notion-${blockColor}`,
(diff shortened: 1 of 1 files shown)


pbteja1998:
I don't think it's valid HTML to have any other html elements inside p tag. There cannot be other p tag inside a p tag.

MartinXPN:
Thanks for the comment @pbteja1998 !
I'm not sure about that but I've seen elements inside `<p>` tags that are rendered properly. An example is: https://www.startupschool.org/posts/47123

I've also tested this on several notion pages and all of them rendered properly.

pbteja1998:
I think browsers are smart enough to render everything properly.

But if we follow html spec, then there can't be a p element inside another p element. But again, it doesn't matter, browsers will render everything properly. But ideally, the leaf block element should be p. There can be other inline elements inside p tag though. Just FYI. We don't have to follow the spec.

https://stackoverflow.com/a/9852381

PR #348, opened 2022-08-11
fix: missing sub-blocks in quote block

zgq354:
#### Description
Quote block can include several sub-blocks, but the render result only renders the first paragraph.

original:  
image

render result:  
image

#### Notion Test Page ID
d3e0c32cb40f4506a024a0672b40ae80


transitive-bullshit (approved):

transitive-bullshit:
Thanks! 🙏

transitive-bullshit:
react-notion-x@6.13.2

Issue #351, opened 2022-08-13
defaultPageCoverPosition is ignored if the notion page used to have a cover


FranciscoMoretti:
#### Description

Notion pages seem to keep the `page_cover_position` in a page `format` even if the cover was removed.
The `Block` component from `packages/react-notion-x/src/block.tsx` uses the `page_cover_position` from the page `format` if it's available instead of the `defaultPageCoverPosition`.

The problem happens when using `NotionRenderer` with `defaultPageCover` and `defaultPageCoverPosition` with a page that used to have a cover. 
In this case it uses the `page_cover_position` from the cover that was removed with the `defaultPageCover`.  `defaultPageCoverPosition` is ignored. 

#### Notion Test Page ID

[This page](https://pewter-piccolo-443.notion.site/Francisco-Moretti-9f7259c2c03f44e9b03b02a9847cf0bf) used to have a cover and has a numeric `page_cover_position` value as part of its format. It doesn't have `page_cover` as part of its format.

transitive-bullshit:
Great catch 💯

PR #352, opened 2022-08-13
Remove page cover position if cover doesn't exist (#351)

FranciscoMoretti:
#### Description

Remove the `page_cover_position` if the `page_cover` doesn't exist in a block format. This allows the block to use the `defaultPageCoverPosition` for `defaultPageCover`.

#351 explains that a cover  `page_cover_position` can exist for a page that doesn't have a `page_cover` if the cover was removed.
> Notion pages seem to keep the `page_cover_position` in a page `format` even if the cover was removed.
The `Block` component from `packages/react-notion-x/src/block.tsx` uses the `page_cover_position` from the page `format` if it's available instead of the `defaultPageCoverPosition`.

>The problem happens when using `NotionRenderer` with `defaultPageCover` and `defaultPageCoverPosition` with a page that used to have a cover. 
In this case it uses the `page_cover_position` from the cover that was removed with the `defaultPageCover`.  `defaultPageCoverPosition` is ignored. 

#### Notion Test Page ID

[This page](https://pewter-piccolo-443.notion.site/Francisco-Moretti-9f7259c2c03f44e9b03b02a9847cf0bf) used to have a cover and has a numeric page_cover_position value as part of its format. It doesn't have page_cover as part of its format.

### DIFF
--- a/examples/full/components/NotionPage.tsx
+++ b/examples/full/components/NotionPage.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react'
 import dynamic from 'next/dynamic'
 import Head from 'next/head'
-import Image from 'next/image'
+// import Image from 'next/image'
 import Link from 'next/link'
 import { useRouter } from 'next/router'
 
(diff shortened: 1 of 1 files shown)


FranciscoMoretti:
FYI @transitive-bullshit this is ready to merge

PR #362, opened 2022-09-01
Fixed groupBy board_coumns being added to non board collections query

normdoow:
#### Description

Apparently  some collections can have some format data on board_columns_by even  when  they aren't board collelctions. This would mess up the payload for our collection query. and  end  up in retrieving no collection rows.

#### Notion Test Page ID

Found the issue on this page originally.  a0c071336c77426ea9925a5088d92ea0
Tested on 2fea615a97a7401c81be486e4eec2e94  and all the collections from the test workspace to makes sure everything works correctly.

### DIFF
--- a/packages/notion-client/src/notion-api.ts
+++ b/packages/notion-client/src/notion-api.ts
@@ -328,9 +328,9 @@ export class NotionAPI {
   ) {
     const type = collectionView?.type
     const isBoardType = type === 'board'
-    const groupBy =
-      collectionView?.format?.board_columns_by ||
-      collectionView?.format?.collection_group_by
+    const groupBy = isBoardType
+      ? collectionView?.format?.board_columns_by
+      : collectionView?.format?.collection_group_by
 
     let filters = []
     if (collectionView.format?.property_filters) {
--- a/packages/notion-client/src/notion-api.ts
+++ b/packages/notion-client/src/notion-api.ts
@@ -344,6 +344,10 @@ export class NotionAPI {
       })
     }
 
+    if (collectionView?.query2?.filter?.filters) {
+      filters.push(...collectionView.query2.filter.filters)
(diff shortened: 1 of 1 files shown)


pbteja1998:
Another quick patch you may want to add is:

https://github.com/pbteja1998/react-notion-x/blob/8c30f73786bf0b89d34fe9dfa0057103bf8f6c45/packages/notion-client/src/notion-api.ts#L348

Check lines 348, 349, 350 in the above file.

If a gallery is filtered by a formula property, react-notion-x is not applying that filter. 

This patch seems to fix that bug.

normdoow:
> Another quick patch you may want to add is:
> 
> https://github.com/pbteja1998/react-notion-x/blob/8c30f73786bf0b89d34fe9dfa0057103bf8f6c45/packages/notion-client/src/notion-api.ts#L348
> 
> Check lines 348, 349, 350 in the above file.
> 
> If a gallery is filtered by a formula property, react-notion-x is not applying that filter.
> 
> This patch seems to fix that bug.

Just added. thanks!

pbteja1998:
https://github.com/pbteja1998/react-notion-x/blob/8c30f73786bf0b89d34fe9dfa0057103bf8f6c45/packages/react-notion-x/src/third-party/property.tsx#L63

(Check Line 63 to Line 92)

This patch can probably go in another PR. Or may be you can just mix it in this PR.

If a gallery has date properties, they are not getting rendered. Looks like we are just trying to render it with `<Text />` which doesn't work.

This patch fixes that and renders the `date property` properly.

normdoow:
> 

ok, will  check that out and add in another PR. or if you want to make a PR for it, id check it out.

pbteja1998:
Awesome. I will make a separate PR then.

PR #365, opened 2022-09-05
fix redeclaration of variable in formula calculation

utkuturunc:
#### Description

let keyword redefined the variable, which set the return value of the entire function to undefined, discarding the calculated value

#### Notion Test Page ID

067dd719a912471ea9a3ac10710e7fdf


utkuturunc:
@transitive-bullshit could you check this one please? a lot of code goes unused because this bug just discards the result

transitive-bullshit:
@utkuturunc looks good. Great catch!

transitive-bullshit:
``react-notion-x@6.15.7`

PR #368, opened 2022-09-08
fix: 🐛 add missing property cover

futantan:

### DIFF
--- a/packages/notion-types/src/collection.ts
+++ b/packages/notion-types/src/collection.ts
@@ -39,6 +39,7 @@ export interface Collection {
   copied_from: string
   template_pages?: Array<ID>
 
+  cover?: string;
   format?: {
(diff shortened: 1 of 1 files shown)


normdoow:
@futantan, thanks for this. Could you  add a test page id where you are seeing an issue with this? Where are you needing the cover type?

packages/notion-types/src/collection.ts
```diff
@@ -39,6 +39,7 @@ export interface Collection {
   copied_from: string
   template_pages?: Array<ID>
 
+  cover?: string;
```
normdoow, line 42: Looks like you are just adding the cover type. More work probably needs to be done to add this in code.

futantan:
Not needed. The data is already there from notion’s api

PR #375, opened 2022-09-22
Add missing optional chaining in getCollectionData function

Asifm95:
#### Description

Since `collectionView` is an optional arg in `getCollectionData()`, calling it without would throw an error because of a missing nullish check.

#### Notion Test Page ID
86b15a89dd4846699383be4476beafa9


normdoow (approved):
Looks good. thanks!

PR #382, opened 2022-10-03
Add support to page_font at BasePageBlock

crisdias:


crisdias:
So we can access if Notion page has font set as "Serif" or "Mono".

image

packages/notion-types/src/block.ts
```diff
@@ -156,6 +156,7 @@ export interface BasePageBlock extends BaseBlock {
     page_cover?: string
     page_icon?: string
     block_color?: Color
+    page_font?: string
```
normdoow, line 159: This just adds the type. I believe we will have to add more logic to take the data and add it to text  throughout a page.

PR #384, opened 2022-10-04
added new tabs for multiple views

normdoow:
#### Description

Notion updated how their databases look  like with multiple views. More info here: https://github.com/NotionX/react-notion-x/issues/383
Updated the dropdown to look more like Notion does currently. As of now,  when there are too many tabs it breaks and goes to the next line.  Not exactly like Notion  but I  think looks fairly good.

image

#### Notion Test Page ID

2fea615a97a7401c81be486e4eec2e94


normdoow:
live react-notion-x v6.14.0

PR #394, opened 2022-10-17
Switch from got to ky + fetch in notion-client

transitive-bullshit:
This PR removes `got` in favor of `ky` in `notion-client`. This means that `notion-client` is no longer tied to Node.js.

[ky](https://github.com/sindresorhus/ky) is a thin wrapper around `fetch`, **which makes `notion-client` compatible with edge runtimes such as Cloudflare Workers, Vercel edge functions, and Deno**.

It also adds a secondary entrypoint to `notion-client` which will only be used by "browsers" (and this should include edge runtimes, though this hasn't been tested yet in practice).

The node.js entrypoint bundles `ky-universal`, which polyfills `node-fetch` and `abort-controller` for node.js so that `ky` works in node.js the same as any other `fetch`-supported environment.

The main advantage of using `ky` instead of `fetch` directly is that `ky` includes robustness features such as automatic HTTP status code-aware retrying.

The main downside of this change is that `got` is more robust on node.js than any other `fetch`-based HTTP client that I'm aware of, including `ky` (though `ky` is the best `fetch`-based HTTP client that I'm aware of).

See @remorses previous PR which started this work in #312 and related discussion in #160

cc @pbteja1998 @MartinXPN @on2air


pbteja1998:
This is awesome. Thanks for this. I will be happy to test it out on Cloudflare Workers once it's live.

normdoow (approved):
Looking  good! This is really cool.

nijamaDev:
Hi, I wanted to know if there was any reason this PR isn't merge yet?

lhojun:
is there any reasons that PR isn't merge?

Issue #411, opened 2022-12-05
PDF content display bug


SwwweetOrange:
#### Description

The PDF content doesn't display from the first page.Official react-notion demo with pdf doesn't display from the frist page and I can't scroll it to the first page.
[Offical demo](https://react-notion-x-demo.transitivebullsh.it/34d650c65da34f888335dbd3ddd141dc)
<img width="1439" alt="截屏2022-12-06 01 55 30" src="https://user-images.githubusercontent.com/71168966/205708050-a9148d33-1df6-4955-8e0e-3b79a625341d.png">

I find it caused by the _display_ of  the pdf asset, I  will create a PR to solve this bug.

PR #412, opened 2022-12-05
fix pdf styling

liqun98:
#### Description

The PDF content doesn't display from the first page.Official react-notion demo with pdf doesn't display from the frist page and I can't scroll it to the first page.
image

It caused by the _display: flex_ at pdf asset, so I change this to _block_.
Solve the issue #411


transitive-bullshit:
Thanks @liqun98 🙏

PR #421, opened 2023-01-16
Fix: notion toggle's child Element overflow issue

JaeSang1998:
#### Description
Hi, I found an issue in Notion Toggle that the child element goes over toggle width when using the nootion link embed. I simply applied the width 100% through the override, and the child element did not overflow, so I'm PRing if it's applicable.

image


jacksongoode:
Good catch!

transitive-bullshit:
Thanks @JaeSang1998 🙏

PR #430, opened 2023-02-04
Hot fix -  new image src with file.notion.so not working

normdoow:
#### Description

Notion made a change where some pages use file.notion.so as their signed urls.
These srcs aren't working currently.
This is a hot fix to at least get it working.
Using the block image src instead just in this case.

#### Notion Test Page ID
d9cc0ffee90e41c4b3ab1f450c346ba7
Don't have a ton of examples to play around with because my Notion workspace isn't producing the file.notion.so srcs. 😅

### DIFF
--- a/packages/notion-utils/src/get-page-image-urls.ts
+++ b/packages/notion-utils/src/get-page-image-urls.ts
@@ -23,7 +23,10 @@ export const getPageImageUrls = (
       if (block) {
         if (block.type === 'image') {
           const signedUrl = recordMap.signed_urls?.[block.id]
-          const source = signedUrl || block.properties?.source?.[0]?.[0]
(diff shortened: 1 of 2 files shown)


pbteja1998:
There is also a file at `packages/notion-utils/src/get-page-image-urls` – not sure where exactly that function is being used, but I think even that should be changed in a similar manner.

normdoow:
> There is also a file at `packages/notion-utils/src/get-page-image-urls` – not sure where exactly that function is being used, but I think even that should be changed in a similar manner.

Yep, I see that. I'll update and merge once I get back to my computer.

nik32co:
@normdoow images are working fine now. However, the videos are still not working. is there any fix coming up for this?

PR #433, opened 2023-02-09
feat: support for normalize japanese hiragana and katakana title

mosugi:
#### Description

I love using react-notion-x on my blog. Thanks again. 
I would like to add Japanese hiragana and katakana to solve the problem similar to #176.

Japanese is written in a mixture of kanji, hiragana and katakana. 
Example: https://jsfiddle.net/uw5m9kts/6/

In [RFC-1738](https://www.ietf.org/rfc/rfc1738.txt), It basically supports us-ascii sets, browsers will encode japanese charaters automatically too.


transitive-bullshit (approved):

transitive-bullshit:
Thanks @mosugi 🙏

Issue #429, opened 2023-02-03
Move to pnpm + turbo


FranciscoMoretti:
#### Description
Just wondering if there is any interest in moving from `yarn` + `lerna` to `pnpm` + `turbo`. 

I've migrated this for myself on a local build and I'm experiencing a better developer experience. 
The benefits I've got from it are:
- Faster installs and dev runs
- Better documentation (`pnpm` + `turbo`)
- Fewer commands. E.g. This command starts dev for the minimal example and the packages that it depends upon: `pnpm dev --filter notion-x-example-minimal...`

My main motivation: As someone who started learning web development recently, migrating to `pnpm` + `turbo` was easier than learning `yarn` + `lerna`. 

I can open a PR with the basic configuration if there is some interest in the topic. Feel free to close the issue if not.

mustaqimarifin:
had the same idea messing with next 13 and a lot of things werent playing too nice. Made small changes to the lazy image wrapper which I thought wouldnt be too bad but I kept having trouble building the project. Anyways, tried out the same combo and its working quite nicely [react-notion-x-turbo](https://github.com/mustaqimarifin/react-notion-x)

transitive-bullshit:
I'm open to this; will make for a tough PR, though.

FranciscoMoretti:
PR Created https://github.com/NotionX/react-notion-x/pull/461. You were right, it's a large one.

PR #461, opened 2023-04-08
#429 Migrate to turbo and pnpm

FranciscoMoretti:
### Description
Replaced `Lerna` and `Yarn` with `Turbo` and `Pnpm`.

Changes:
- More `tsup` and less `tsc`
  - `tsup` is the recommended tool in [`turbo` docs](https://turbo.build/repo/docs/handbook/publishing-packages/bundling)
  - `tsup --watch` replaces all the `watch` tasks. It automatically re-builds a package when in development tasks. [Docs](https://tsup.egoist.dev/#watch-mode)
  - `tsup` now generates the type definitions with the config `dts: true` on `tsup.config.ts` [Docs](https://tsup.egoist.dev/#generate-declaration-file)
- tsconfig's `composite` is not supported by `tsup`, so I've changed it to `false` https://github.com/egoist/tsup/issues/571.
  - This is also the recommended config on the [`turbo` docs](https://turbo.build/repo/docs/handbook/linting/typescript) 
- dependencies are linked internally with `"workspace:*"`. They should get replaced by the actual version when publishing [Pnpm workspaces](https://pnpm.io/workspaces).
- Documentation updates (Readme files)
- Organized tasks to match turborepo pattern
  - This pattern parallelizes tasks and only runs what's needed [`turbo` tasks docs](https://turbo.build/repo/docs/core-concepts/monorepos/running-tasks)
  - Lint tasks moved to each workspace. [`turbo` linting docs](https://turbo.build/repo/docs/handbook/linting)
  - Test task dependencies coordinated through turbo pipeline (`turbo.json`)
  - "pre" tasks are defined in pipeline as dependencies. Therefore, they were removed from `package.json`
  - publish tasks moved to workspaces and dependencies defined in `turbo.json`.
    - Defined dependencies only for workspaces that use publish because of this bug https://github.com/vercel/turbo/issues/2988
- Renamed `deploy` task to `vercel-deploy` in examples to avoid confusion with `pnpm deploy`
  - Now running `pnpm vercel-deploy` and `pnpm run vercel-deploy` are valid from an example directory 
  - The alternative is to keep the `deploy` name but that would only work with `pnpm run deploy` because `pnpm deploy` is an existent pnpm command.
- Configured tasks caches with [`turbo` caching](https://turbo.build/repo/docs/core-concepts/caching).
 

#### Benefits
This change has the following advantages:
- Simplifies processes: E.g. You can now run `dev` with a single command
- Faster re-builds: Turbo auto caches builds per package
- Faster builds, lint, test: Turbo runs things in parallel whenever possible
- Faster re-installs: Pnpm auto caches node modules
- Modern tools: Using these new tools is a better developer experience overall and they have better documentation.

#### Turbo Examples
The configuration was inspired by Turborepo official examples
- Basic example: https://github.com/vercel/turbo/tree/main/examples/basic
- CRA example: https://github.com/vercel/turbo/tree/main/examples/with-create-react-app
- Kitchen sink example; https://github.com/vercel/turbo/tree/main/examples/kitchen-sink

- This bugfix was needed for a correct build of `notion-x-example-full` https://github.com/NotionX/react-notion-x/pull/459

### Tests
[x] Installing
```
pnpm i
```

[x] Running a global build:
```
turbo build
```
[x] Running a build from the package dir
```
cd packages/notion-utils
pnpm build
```
[x] Running every dev task
```
turbo dev
```
[x] Running tests execute `unit` `eslint` and `prettier`.
```
turbo test
```

[x] Running dev in a single example, modifying a dependency, and verifying it rebuilds it
```
turbo dev --filter=notion-x-example-minimal...
```

[x] Connecting with `nextjs-notion-starter-kit`.
  - `yarn` link and the contributing docs from that repo are still valid after this change https://github.com/transitive-bullshit/nextjs-notion-starter-kit/blob/main/contributing.md

[x] Running publish executed the correct tasks
```
turbo publish
```

[x] Running vercel deploy
```
cd examples/minimal
pnpm run vercel-deploy
``` 

[x] The cache works as expected in all tasks


FranciscoMoretti:
Feedback is encouraged. I had to make a few choices here and there and I'm not sure if they are the best ones.

Also, it seems like deployments are failing and I can't inspect or check details. @transitive-bullshit let me know if there is anything I can do to fix them.

transitive-bullshit:
This is great @FranciscoMoretti && sorry I didn't get to this sooner.

I'm currently doing a maintenance sweep which includes these changes.

PR #470, opened 2023-04-19
feat:support for normalize korean title

woensug-choi:
#### Description

Just like adding Chinese (https://github.com/NotionX/react-notion-x/pull/177) and Japanse (https://github.com/NotionX/react-notion-x/pull/433), fix for Korean titles.

For those who are facing error for their own languages, use `includeNotionIdInUrls: true` in the `site.config.ts` to bypass the issue.


transitive-bullshit:
Thanks!

PR #473, opened 2023-04-23
feat: support reducer result

maxswjeon:
#### Description

There is a `reducer` type of collection view, which is `table` type when created on Notion Website.

#### Notion Test Page ID
d297b76916ff4749b49648ac9984c358


maxswjeon:
I don't know why the test fails... I cannot see the Vercel logs

maxswjeon:
Found it. ReducerCollectionView should extend BaseCollectionView

transitive-bullshit:
Thanks!

PR #477, opened 2023-04-29
fix: empty group title for board view

junogueira:
#### Description

Fix title for empty group on board view, using the property name instead of a static title.

#### Example

Board view grouped by **Priority** property:

Before
Screenshot 2023-04-29 at 19 56 03

After
Screenshot 2023-04-29 at 19 55 43

#### Notion Test Page ID

`16f256fbf9b54bc6b4befe13ba74faeb`

### DIFF
--- a/packages/react-notion-x/src/third-party/collection-view-board.tsx
+++ b/packages/react-notion-x/src/third-party/collection-view-board.tsx
@@ -66,8 +66,6 @@ function Board({ collectionView, collectionData, collection, padding }) {
     collectionView?.format?.board_groups2 ||
     []
 
-  const boardGroupBy = collectionView?.format?.board_columns_by?.groupBy
-
   const boardStyle = React.useMemo(
     () => ({
       paddingLeft: padding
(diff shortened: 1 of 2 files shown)


packages/react-notion-x/src/third-party/collection-view-board.tsx
```diff
@@ -109,8 +109,8 @@ function Board({ collectionView, collectionData, collection, padding }) {
                       />
                     ) : (
                       <span>
-                        <EmptyIcon className='notion-board-th-empty' /> No
-                        Select
+                        <EmptyIcon className='notion-board-th-empty' />
+                        {`No ${schema.name}` || 'No Select'}
```
s4nju, line 113: Here you are missing the check for `schema.name`, what this will render if schema.name is not available?

Your provided `No Select` optional fallback does not work here.

transitive-bullshit: agreed

junogueira: Thanks @s4nju and @transitive-bullshit, it should be fixed now.

transitive-bullshit:
Thanks!

PR #481, opened 2023-05-11
feat: support append node after head like account avatar etc.

rowthan:

### DIFF
--- a/packages/react-notion-x/src/components/header.tsx
+++ b/packages/react-notion-x/src/components/header.tsx
@@ -13,12 +13,14 @@ import { SearchDialog } from './search-dialog'
 
 export const Header: React.FC<{
   block: types.CollectionViewPageBlock | types.PageBlock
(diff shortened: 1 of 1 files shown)


packages/react-notion-x/src/components/header.tsx
```diff
@@ -13,12 +13,14 @@ import { SearchDialog } from './search-dialog'
...
   block: types.CollectionViewPageBlock | types.PageBlock
-}> = ({ block }) => {
+  children?: React.ReactNode
+}> = ({ block,children }) => {
   return (
     <header className='notion-header'>
       <div className='notion-nav-header'>
         <Breadcrumbs block={block} />
         <Search block={block} />
+        {children}
```
s4nju, line 23: This will not work as, the container className have `justify-content: space-between` which will push the Search button in the middle.

If you wanted to give placeholder for avatar this is not the correct solution.

PR #491, opened 2023-05-27
Fix block equation display issue

drjackchan:
#### Description

The inline variable is hardcoded to false at the moment, making all the equations rendered to be inline equations. Fixing this issue by removing the hardcoding. 

#### Notion Test Page ID

39a20e59ea4147f1b3e145166528ab95

### DIFF
--- a/examples/cra/package.json
+++ b/examples/cra/package.json
@@ -1,6 +1,6 @@
 {
   "name": "notion-x-example-cra",
-  "version": "6.15.8",
+  "version": "6.16.0",
   "private": true,
   "scripts": {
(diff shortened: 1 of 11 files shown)


jacksongoode:
Is this not addressed in https://github.com/NotionX/react-notion-x/pull/281?

jackccy:
I looked at change #281 , it seems to me that it is hardcoding the display mode != inline? How can that differentiate inline and block equation?

transitive-bullshit:
I included the core of the fix for `block` but there are other changes in this PR that are unrelated, so closing.

PR #494, opened 2023-06-04
Fix: setting app-element in react-modal Modal component

s4nju:
#### Description

Added app-element to the react-modal, earlier it was giving warning, which is now fixed in this PR.


transitive-bullshit:
Thanks! 🙏

PR #499, opened 2023-06-14
hotfix / remove log

stefanfeldner:
@transitive-bullshit Can we get rid of this console.log() please, it clutters the logs.

Also, no TS types are available for `"notion-compat": "^6.16.0",` version.
There is also no `@types/notion-compat`, so TS complains when importing it, and I have to ts-expect-error.

image


konhi (approved):

transitive-bullshit:
Thanks! 🙏

PR #514, opened 2023-08-25
skip pages which is not alive.

ulion:
#### Description
when looping get all pages, it's possible that encountered alive===false block/page, it should be skipped.


transitive-bullshit:
Thanks! 🙏

PR #523, opened 2023-10-03
trigger build and publish

normdoow:
#### Description

Needed a new version to push the latest to npm.


qiyuan-wang:
npm package looks like keep in v6.16.0.

PR #528, opened 2023-10-30
Custom components

oleksiisurnin:
#### Description

I've implemented a feature that allows the creation of custom components for each block, such as Assets, Header, Text, Table, and others. This functionality enables users to completely override individual components as needed. Personally, I will be using this in my project for more refined SEO settings, styles and content CDN adjustments.

#### Notion Test Page ID

For testing purposes, you can use the standard "minimal" project.


transitive-bullshit:
I really like this idea in theory, but it's too far out-of-date in practice.

PR #533, opened 2023-12-11
fix: add missing table_block_row_header to TableBlock type

ahonn:
#### Description

Add missing table_block_row_header to TableBlock type.
When setting the first row of the table as the header, the table block has this field.


ahonn:
@normdoow Can this PR be merged? It seems like there is no one currently handling the PRs in this repo.

rimonhanna:
@transitive-bullshit

PR #563, opened 2024-10-03
feat: add `card_cover_position` to collection card

mfts:


mfts:
You're quick @transitive-bullshit 


Thought you're not checking here anymore  😍

PR #567, opened 2024-10-30
Add missing new attribute 'card_cover_position' to BasePageBlock type

onnimonni:
When the PR #563 was merged the types were not updated and this broke the builds and typing.

After this is added the builds should work again.


transitive-bullshit:
Thanks @onnimonni 🙏

PR #568, opened 2024-10-30
Add and document mise (asdf clone) to enable reproducible environment for the project

onnimonni:
Getting the project up and running with all exotic aging dependencies is pretty hard. This is my best effort to make it easier for everyone who is contributing.

After merging this users can get working environment by using [mise](https://mise.jdx.dev/) or [asdf](https://asdf-vm.com/).

Mise is preferred because it can also automatically install working version of yarn.

PR #573, opened 2024-10-31
Update node version to 18.X to fix ky universal build issues

onnimonni:
## Context
This PR builds on top of #568. That PR should be merged first so that we get smaller overall changes.

## Root cause
After @transitive-bullshit replaced `got` for `ky` #394 our builds broke down with following issue:
```
Error: R] Could not resolve "ky-universal"

    src/notion-api-universal.ts:2:7:
      2 │ import 'ky-universal'
        ╵        ~~~~~~~~~~~~~~

  You can mark the path "ky-universal" as external to exclude it from the bundle, which will remove this error.
```

I noticed that the `ky-universal` is not needed anymore if we update `ky` to `1.0.0` and nodejs to version `18.X` version.

## Consequences and why we need to add tsimp

After I upgraded them our tests started to fail in all packages:
```
lerna ERR! yarn run test exited 1 in 'notion-compat'
lerna ERR! yarn run test stdout:
$ ava

  Uncaught exception in src/notion-compat-api.test.ts

  TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for /Users/onnimonni/Projects/react-notion-x/packages/notion-compat/src/notion-compat-api.test.ts

  ✖ src/notion-compat-api.test.ts exited with a non-zero exit code: 1
  ─

  1 uncaught exception
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
lerna ERR! yarn run test stderr:
error Command failed with exit code 1.
lerna ERR! yarn run test exited 1 in 'notion-compat'
```

I wasn't able to figure out this on my own but I was using the [cursor editor](https://www.cursor.com/) and it suggested adding `tsimp` as loader for `ava` and then the tests started to work.

We can't use the latest version of `tsimp` so it's pinned to 2.0.11.
It can be updated when we get node 18 support.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,11 +6,11 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
-      - uses: actions/setup-node@v3
+      - uses: actions/setup-node@v4
         with:
-          node-version: 16
+          node-version-file: .tool-versions
           cache: yarn
 
       - run: yarn install --frozen-lockfile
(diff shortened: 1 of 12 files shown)


transitive-bullshit:
Fixed in https://github.com/NotionX/react-notion-x/pull/571

PR #586, opened 2024-11-18
Fix callout spacing for new callouts (sans icons)

wustep:
#### Description

See issue https://github.com/NotionX/react-notion-x/issues/565

Notion recently introduced two changes to callouts:
- There is no longer a forced text title and instead we just have block contents, allowing blocks like `<h1>` at the top of a callout.
- You can remove the icon from the callout entirely (this was me 😉).

This led to some minor visual issues with react-notion-x.

In this PR, we: 
- Remove the icon entirely when it is missing 
- Removes the top padding/margins on the callout text's first child.

#### Notion Test Page ID

1425cb08cf2c80efb771e6010d8dac21

**Notion**
CleanShot 2024-11-18 at 01 07 33@2x

**Main [before change]**
https://react-notion-x-demo.transitivebullsh.it/1425cb08cf2c80efb771e6010d8dac21

CleanShot 2024-11-18 at 01 07 19@2x

**After change**
CleanShot 2024-11-18 at 01 09 22@2x

Old callouts are fine too
CleanShot 2024-11-18 at 01 28 57@2x

(Note: there seems to be a funny bug with adjacent headings being rendered on the same line 😛. This is not addressed in this PR.)

### DIFF
--- a/packages/react-notion-x/src/block.tsx
+++ b/packages/react-notion-x/src/block.tsx
@@ -838,5 +838,4 @@ export function Block(props: BlockProps) {
 
       return <div />
   }
-
 }
--- a/packages/react-notion-x/src/block.tsx
+++ b/packages/react-notion-x/src/block.tsx
@@ -592,7 +592,7 @@ export function Block(props: BlockProps) {
               blockId
             )}
           >
-            <PageIcon block={block} />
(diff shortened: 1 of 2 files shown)


transitive-bullshit:
Thanks @wustep 🙏

PR #590, opened 2024-11-21
fix: load images by safely add spaceId to searchParams

### LINKED ISSUE #588: Some tweets & images are not loading
#### Description
(repros in both `react-notion-x` and `nextjs-notion-starter-kit`)

#### Tweets

CleanShot 2024-11-19 at 02 18 39@2x

CleanShot 2024-11-19 at 01 04 02@2x

```
function Tweet({ id }: { id: string }) {
  const { recordMap } = useNotionContext()
  const tweet = (recordMap as types.ExtendedTweetRecordMap)?.tweets?.[id]

  return (
    <React.Suspense fallback={<TweetSkeleton />}>
      {tweet ? <EmbeddedTweet tweet={tweet} /> : <TweetNotFound />}
    </React.Suspense>
  )
}
```

This doesn't work, because `id` is actually `1772670594779079043&spaceId=30725683-e071-41f1-988d-e6e6fa72abd8`
but the recordMap stores `1772670594779079043` as the key.

Easy workaround is just to ignore the `&spaceId=...`

#### Images

Similarly, we see an issue with images as well -- see comments.

#### Notion Test Page ID
3b052104657a43feaf2230c2eb2a4036

mfts:
#### Description

This PR fixes the broken images and files in Notion. 

Before: spaceId was just added with and `&` even if it was the first query params.
Now: spaceId is added safely to the searchParams using URL operations. 

fixes #588

### DIFF
--- a/packages/react-notion-x/src/components/asset.tsx
+++ b/packages/react-notion-x/src/components/asset.tsx
@@ -135,7 +135,9 @@ export function Asset({
   }
 
   if (block.space_id) {
-    source = source.concat('&spaceId=', block.space_id)
+    const url = new URL(source)
+    url.searchParams.set('spaceId', block.space_id)
+    source = url.toString()
   }
 
   let content = null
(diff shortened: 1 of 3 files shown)


transitive-bullshit:
Thanks @mfts 🙏

PR #592, opened 2024-11-25
Add support for Table columns & headers

wustep:
#### Description

Right now, react-notion-x assumes that the first row is always the header, but in Notion you can customize either:
- no header at all
- 1st row is header
- 1st column is header
- 1st row and first column are headers

This adds support for these, removing the default first-row-as-header.

#### Notion Test Page ID

https://www.notion.so/wustep/Tables-1495cb08cf2c804f977af6108cbd4fc8

CleanShot 2024-11-25 at 15 33 31@2x


wustep:
@transitive-bullshit ping ;)

PR #596, opened 2024-12-14
Fix private file signed url generation

worgho2:
#### Description

When trying to display a PDF file, I've figured out that notion is using a new bucket url to serve their files and it impacts how this library handles url signing. 

- Old files start with: `https://s3-us-west-2.amazonaws.com/secure.notion-static.com/...`
- New files start with: `https://prod-files-secure.s3.us-west-2.amazonaws.com/...`

Including `prod-files-secure` prefix when identifying a private url, fixed the issue.

Other than that, the PDF component is using an outdated `pdfjs-dist` cdn url, so i've updated it.

#### Notion Test Page ID

- PDFs not being displayed: `34d650c65da34f888335dbd3ddd141dc`

### DIFF
--- a/packages/notion-client/src/notion-api.ts
+++ b/packages/notion-client/src/notion-api.ts
@@ -237,17 +237,17 @@ export class NotionAPI {
         // console.log(block, source)
 
         if (source) {
-          if (!source.includes('secure.notion-static.com')) {
-            return []
+          if (source.includes('secure.notion-static.com') || source.includes('prod-files-secure')) {
+            return {
+              permissionRecord: {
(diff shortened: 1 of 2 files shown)


scriptify:
Good stuff, thanks

transitive-bullshit:
Thanks @worgho2 🙏

mfts:
@transitive-bullshit this is not yet published a new version, right?

transitive-bullshit:
> @transitive-bullshit this is not yet published a new version, right?

Correct; will publish soon.

PR #602, opened 2025-01-11
Replaced third-party lazy-loading implementation with native lazy-loading for improved performance

rodriguescarson:
### Performance Benefits:
- **Reduced JavaScript Overhead**: Native lazy-loading is handled by the browser, eliminating the need for additional libraries and reducing the initial JS load.
- **Optimized by the Browser**: Modern browsers automatically optimize native lazy-loading, resulting in better performance and lower memory usage.
- **Improved Initial Load**: By leveraging native lazy-loading, we reduce the number of unnecessary requests, improving page load times, especially on slower connections or devices.
- **No Extra Dependencies**: This change removes the need for third-party libraries like `react-lazy-images` or `react-lazy-load-image-component`, making the codebase simpler and more maintainable.

This change leverages the built-in `loading="lazy"` attribute in React for images, which provides a significant performance boost by deferring image loading until they're close to the viewport, optimizing the user experience on initial load.

### DIFF
--- a/examples/full/next-env.d.ts
+++ b/examples/full/next-env.d.ts
@@ -2,4 +2,4 @@
 /// <reference types="next/image-types/global" />
 
 // NOTE: This file should not be edited
-// see https://nextjs.org/docs/pages/building-your-application/configuring/typescript for more information.
+// see https://nextjs.org/docs/pages/api-reference/config/typescript for more information.
--- a/examples/minimal/next-env.d.ts
+++ b/examples/minimal/next-env.d.ts
@@ -2,4 +2,4 @@
 /// <reference types="next/image-types/global" />
 
 // NOTE: This file should not be edited
-// see https://nextjs.org/docs/pages/building-your-application/configuring/typescript for more information.
+// see https://nextjs.org/docs/pages/api-reference/config/typescript for more information.
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "@fisch0920/eslint-config": "^1.4.0",
     "@total-typescript/ts-reset": "^0.6.1",
     "@types/node": "^22.9.0",
+    "@types/react-lazyload": "^3.2.3",
     "del-cli": "^6.0.0",
     "eslint": "^8.57.1",
     "npm-run-all2": "^7.0.1",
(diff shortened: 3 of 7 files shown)


rodriguescarson:
Hello Mainter, can you let me know the problem, why the build failed?

rodriguescarson:
## Replaced third-party lazy-loading implementation with native lazy-loading for improved performance.

### Performance Benefits:
- **Reduced JavaScript Overhead**: Native lazy-loading is handled by the browser, eliminating the need for additional libraries and reducing the initial JS load.
- **Optimized by the Browser**: Modern browsers automatically optimize native lazy-loading, resulting in better performance and lower memory usage.
- **Improved Initial Load**: By leveraging native lazy-loading, we reduce the number of unnecessary requests, improving page load times, especially on slower connections or devices.
- **No Extra Dependencies**: This change removes the need for third-party libraries like `react-lazy-images` or `react-lazy-load-image-component`, making the codebase simpler and more maintainable.

This change leverages the built-in `loading="lazy"` attribute in React for images, which provides a significant performance boost by deferring image loading until they're close to the viewport, optimizing the user experience on initial load.

rodriguescarson:
It works better!

rodriguescarson:
@NiklasPor apologies for the very late proper reply, and fair questions on both counts.

On `zoomable`: removing it was a mistake. Going back through the diff, this branch cut a lot more than the `react-lazy-images` dependency. It also dropped the medium-zoom attach path, the `preview_images` blur-up placeholders, and the `components.Image` integration, so anyone rendering through `next/image` or relying on preview images would have regressed. That wasn't a deliberate tradeoff I'd thought through, and my one-line answer at the time didn't deserve to pass for one.

You were right about the description too. It was a padded list of generic benefits instead of an explanation of what actually changed and what it cost.

Swapping `react-lazy-images` for native `loading="lazy"` still seems worth doing, but only in a version that keeps zoom, preview images and the custom `Image` component working, and without the unrelated prettier bump and `next-env.d.ts` edits this branch is carrying. This one is 18 months stale and conflicting, so I'm closing it rather than leaving it sitting in your queue. Thanks for taking the time to look at it.

PR #607, opened 2025-02-12
fix(collection): two collection hydration errors

nightohl:
#### Description

- change `useWindowSize` hook to ensure for ssr result to match when hydration.
- fixed two hydration errors caused by `windowWidth` and `padding`

image


transitive-bullshit:
@gemini-code-assist review this PR

PR #608, opened 2025-02-12
fix(code): code hydration error

nightohl:
#### Description

- `prismjs` is under the hood, it addes `language` and `tabIndex` attributes to `pre` tag but the ssr results doesn't have these, so **hydration error** occures.

image


rimonhanna:
@transitive-bullshit

transitive-bullshit:
thank you @nightohl 🙏

PR #609, opened 2025-02-12
feat(text): support Link Mention format

nightohl:
#### Description

image

- Fixed an issue where the newly introduced "Link Mention" feature in Notion was unsupported, causing an "unsupported text format" error.

image

- Now, it's correctly rendered in a way similar to Notion's functionality.


rimonhanna:
@transitive-bullshit

transitive-bullshit:
Thank you @nightohl 🙏

PR #610, opened 2025-02-15
Fix: title key can't exist inside title, rich_text should be the key

harsh26shah03:
#### Description

For notion-compat,

while converting blocks, the child_page typed block have title, to convert that we need to access rich text of that title, byt we are trying to access title.title, so it's undefined and throwing error.


rimonhanna:
@transitive-bullshit

PR #612, opened 2025-03-05
feat: add mention link support

qixing-jk:
#### Description

Fixed an issue where the newly introduced "Link Mention" feature in Notion was unsupported, causing an "unsupported text format" error.
Clip_20250306_005559

It can now be displayed in a manner similar to page link

### DIFF
--- a/packages/notion-types/src/core.ts
+++ b/packages/notion-types/src/core.ts
@@ -119,6 +119,13 @@ export type InlineEquationFormat = ['e', string]
 export type DiscussionFormat = ['m', string]
 export type ExternalLinkFormat = ['‣', [string, string]]
 export type DateFormat = ['d', FormattedDate]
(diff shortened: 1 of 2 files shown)


nightohl:
@qixing-jk 
Hi there! I think we've looked the same issue!

* #609 

is merged so It seems to be kinda duplicated feature.
How about take a look the merged code and enhancing it?

qixing-jk:
> @qixing-jk[](https://github.com/qixing-jk) Hi there! I think we've looked the same issue!
> 
> * [feat(text): support Link Mention format #609](https://github.com/NotionX/react-notion-x/pull/609)
> 
> has been merged so It seems to be kinda duplicated feature. How about take a look the merged code and enhancing it?

I'm happy that that feature was implemented, that PR does look great, I'll try to write better code in the future

PR #613, opened 2025-03-10
fix: look for `attachment:` in file upload source

mfts:
#### Description

Notion file uploads now append `attachment:` to the file identifier instead of the s3 url.

Therefore the source is also a valid source if it includes `attachment:` in the source text.

#### Notion Test Page ID

Here's an example of a Notion page

1b2d3f870f0080caae0bc0b2766cf31f


transitive-bullshit:
Thanks @marc 🙏

PR #614, opened 2025-03-12
fix(eslint): imports sort lint

nightohl:
#### Description

* #609
@transitive-bullshit Thanks for merging my PR!
but It seems I missed running the eslint check for sorting import order. This PR is to fix that so it can pass the merge process.

Also, I'm gonna update the contribution guide and add a lint check to the GitHub Actions for PRs and pushes, and submit a new PR for that as well. 
- [x] Done
* #615


transitive-bullshit:
no worries :)

PR #615, opened 2025-03-12
Add Contribution Guide for PR Process and Lint Test in GitHub Actions

nightohl:
#### Description

We’ve merged the #604 feature into the master branch, but unfortunately, an ESLint error was overlooked during the review process. The issue was addressed and fixed in PR #614.

To prevent similar situations in the future, I would like to suggest updating the contribution guide and GitHub Actions for PRs to include an automated lint check.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,6 +33,9 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile --strict-peer-dependencies
 
+      - name: Run lint
+        run: pnpm run test:lint
+
       - name: Run test
         run: pnpm test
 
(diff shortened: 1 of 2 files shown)


nightohl:
@transitive-bullshit 
It's the code for the ESLint issue that was missed during the PR stage. 
Please take a look :)

transitive-bullshit:
`pnpm test` already runs `test:format` and `test:lint`, so no need to run separate commands.

I just didn't run the tests before merging the other PRs.

nightohl:
@transitive-bullshit 
Yess you're right, test already includes format, lint, etc.
Then I couldn't find why the past one can pass the test..
Thanks for reply :) I'll close this PR

PR #630, opened 2025-05-06
feat: table header rendering optimization

### LINKED ISSUE #628: Table Style
Notion Table provide the following Options: 

Image

In Notion X React however, tables automatically apply a gray background to the first row via default system styling, regardless of this setting.
This leads to unwanted styling, when no header is intended.

LooseLi:
#### Description

Fixes #628 

Simple tables always show the first row with header styling (background color), regardless of whether header settings are enabled. This is inconsistent with Notion's native behavior.

This PR fixes table header rendering to only apply styling when explicitly enabled:

- Row headers (first row): Only styled when `table_block_column_header` is true
- Column headers (first column): Only styled when `table_block_row_header` is true

Note: There's an interesting naming confusion in Notion's API - table_block_column_header actually controls row headers, while table_block_row_header controls column headers. This PR maintains the original field names for API compatibility but implements the correct behavior.

**After**
image

#### Notion Test Page ID

1eb1578febe58070a07edf70bd34025d

### DIFF
--- a/packages/react-notion-x/src/block.tsx
+++ b/packages/react-notion-x/src/block.tsx
@@ -789,6 +789,10 @@ export function Block(props: BlockProps) {
       const formatMap = tableBlock.format?.table_block_column_format
       const backgroundColor = block.format?.block_color
 
+      const hasRowHeader = tableBlock.format?.table_block_column_header === true
+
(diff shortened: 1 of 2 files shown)


transitive-bullshit:
thank you for the high quality PRs @LooseLi 🙏

PR #635, opened 2025-05-27
Fix/support multi relation property

rimonhanna:
#### Description

A database with sub-items enabled 
Add multiple sub-items to a single "parent" page
Open parent page:
Previously: It showed only the first child/sub-item as the returned recordMap only included the full information of the first, and only the Ids of the others
Now: It attempts to fetch all children/sub-items blocks and shows them as expected

### DIFF
--- a/examples/full/lib/notion.ts
+++ b/examples/full/lib/notion.ts
@@ -21,7 +21,7 @@ if (useOfficialNotionAPI) {
 }
 
 export async function getPage(pageId: string): Promise<ExtendedRecordMap> {
-  const recordMap = await notion.getPage(pageId)
+  const recordMap = await notion.getPage(pageId, { fetchRelationPages: true })
 
   if (previewImagesEnabled) {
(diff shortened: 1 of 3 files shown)


transitive-bullshit:
hmmmm the `minimal` example is intentionally minimal and does not support relations / collections.

@rimonhanna does your use case work with the `full` example? that's meant to handle this case.

rimonhanna:
@transitive-bullshit any chance this can be merged and released soon? Thanks!

rimonhanna:
https://reactexamples.notion.site

transitive-bullshit:
awesome; thanks again, rimon 🙏

transitive-bullshit:
ahhh actually, what's the notion id of that page? i can't seem to find it on that site

rimonhanna:
@transitive-bullshit i made it clone-able so you can add it to your demo workspace

PR #640, opened 2025-06-02
fix: support direct image hyperlinks

### LINKED ISSUE #637: Support for hyperlinks attached to images
While using the library, I thought it would be great to have native support for hyperlinks applied directly to images. 
After reviewing [this issue](https://github.com/NotionX/react-notion-x/issues/306), it seems that links can be supported via image captions. 
However, using captions to display hyperlinks feels a bit awkward, as it exposes the link directly within the Notion page.

I believe we could add support for actual image hyperlinks by modifying the asset-wrapper file directly. Here’s a snippet showing how this might be implemented:
```typescript
const caption = block.properties?.caption?.[0]?.[0]
const imageHyperlink = (block as ImageBlock).format?.image_hyperlink

const isURL = getIsURL(block, imageHyperlink, caption)

function getIsURL(block: BaseContentBlock, imageHyperlink?: string, caption?: string): boolean {
  if (block.type !== 'image') {
    return false
  }

  if (imageHyperlink) {
    const id = parsePageId(imageHyperlink, { uuid: true })
    return imageHyperlink.charAt(0) === '/' && !!id
  }

  if (caption) {
    const id = parsePageId(caption, { uuid: true })
    const isPage = caption.charAt(0) === '/' && id
    return !!(isPage || isValidURL(caption))
  }

  return false
}

```

raymondanythings:
fix #637 

#### Description
adds native support for hyperlinks applied directly to image blocks. While Notion allows hyperlinks via image captions, exposing the link as a caption feels intrusive and affects the user experience.

To address this, we modify the asset-wrapper logic to properly detect and wrap images in anchor tags when a valid image_hyperlink is present.

#### Solution

1. **Introduced `getURLInfo()`** 
- Determines whether the image should link to
  - a Notion page (/xxx pattern with UUID),
  - an external URL (validated using a URL regex).

2. **Improved link handling logic**
- Fallback order: imageHyperlink → caption
- Automatically wraps the image with <components.PageLink> only when a valid link is detected.
- Maintains support for internal Notion page routing via `mapPageUrl`.

#### Notion Test Page ID
2061b06afbf1807b8e35d35e7a7fd146


Orchemi:
I really hope this pull request is merged into main soon.

Issue #661, opened 2025-08-21
fix: 'syncRecordValues' endpoint returns 404 error


chugue:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `notion-client@7.4.3` for the project I'm working on.

And I had an issue retrieving a page using the 'syncRecordValues' endpoint, which resulted in a 404 status code.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/notion-client/build/index.js b/node_modules/notion-client/build/index.js
index 2ad4eba..0377544 100644
--- a/node_modules/notion-client/build/index.js
+++ b/node_modules/notion-client/build/index.js
@@ -432,7 +432,7 @@ var NotionAPI = class {
   }
   async getBlocks(blockIds, kyOptions) {
     return this.fetch({
-      endpoint: "syncRecordValues",
+      endpoint: "syncRecordValuesMain",
       body: {
         requests: blockIds.map((blockId) => ({
           // TODO: when to use table 'space' vs 'block'?
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

Issue #659, opened 2025-08-19
Request failed with status code 400 Bad Request: POST https://www.notion.so/api/v3/syncRecordValues


asp3:
Running into this all of a sudden. Maybe notion took down this API?

danielpowell4:
More info

Error type: HTTPError
Error message: Request failed with status code 400 Bad Request: POST https://www.notion.so/api/v3/syncRecordValues
Status: 400
Status text: Bad Request
Request URL: https://www.notion.so/api/v3/syncRecordValues
Response body: "This endpoint doesn't exist anymore! What are you doing? Email team@makenotion.com'

xhemj:
Hello! I've noticed that Notion appears to have changed their API endpoint from `/api/v3/syncRecordValues` to `/api/v3/syncRecordValuesSpaceInitial`. The original endpoint seems to be no longer functional.

Here's an example of the new API call (I'm not sure if this API endpoint existed before):

```
POST https://www.notion.so/api/v3/syncRecordValuesSpaceInitial

[
    {
        "pointer": {
            "table": "block",
            "id": "bc353a0c-xxxx-xxxx-xxxx-xxxxe0b3adfe",
            "spaceId": "14b2d77e-xxxx-xxxx-xxxx-xxxxc3827c80"
        },
        "version": -1
    },
    ...
]
```

nathan-ahn:
Here's a workaround solution:

```typescript
const recordMap = await notion.getPage(pageId, {
	gotOptions: {
		hooks: {
			beforeRequest: [
				(options) => {
					if (options.url.pathname === "/api/v3/syncRecordValues") {
						options.url.pathname = "/api/v3/syncRecordValuesMain";
					}
				},
			],
		},
	},
});
```

I found that `/api/v3/syncRecordValuesSpaceInitial` was returning 500s, so I went with `/api/v3/syncRecordValuesMain` instead.

nnfans:
> I think we should use the `kyOptions` property!
> 
> const notionAPI = new NotionAPI({
>   activeUser: process.env.NOTION_ACTIVE_USER,
>   authToken: process.env.NOTION_TOKEN_V2,
>   kyOptions: {
>     hooks: {
>       beforeRequest: [
>         (request, options) => {
>           const url = request.url.toString();
> 
>           if (url.includes('/api/v3/syncRecordValues')) {
>             return new Request(
[...]

PR #662, opened 2025-08-21
fix: update endpoint in getBlocks method to syncRecordValuesMain

LHDi:
#### Description

`getBlocks` uses `syncRecordValues` endpoint which is now deperecated and no longer works unfortunatly. This PR update that to the working endpoint `syncRecordValuesMain` 
Related to #659 (suggest `beforeRequest` hook) and #661 (suggest package patching)

#### Notion Test Page ID

0fa06897b70f4e78876324cd9df76356


tomcerdeira:
does this actually work? is there a way for me to try it out?

PR #668, opened 2025-10-01
fix: apply list_start_index for numbered_list

ferrariRoma:
#### Description

apply numbered list start index (`list_start_index`) in `getListNumber`.
if the block type is `numbered_list` and `format.list_start_index` exists, use that value.

#### Notion Test Page ID

27fadccd81a280fc8ca5ebab12ce75fe

I leaved some explain at test page

before
image

after
image


ferrariRoma:
@transitive-bullshit review this pr please~

transitive-bullshit:
thanks @ferrariRoma 🙏

PR #687, opened 2026-03-07
style: use CSS variable for global corner radius

albertodainotti:
## Summary

Replaces hardcoded `border-radius: 3px` values throughout `styles.css` with a new CSS custom property `--notion-corner-radius: 8px`.

## Changes

- Added `--notion-corner-radius: 8px` CSS variable to the root `:root` scope
- Replaced 10 instances of `border-radius: 3px` with `var(--notion-corner-radius)`
- Affected elements: page icons, inline code, code blocks, bookmarks, callouts, collection cards, collection card covers, search result items, and asset wrapper icons

## Motivation

This makes the corner radius consistent and easily customizable. Users can override `--notion-corner-radius` in their own stylesheets to adjust the global border radius across all Notion components without needing to target individual selectors.

### DIFF
--- a/packages/react-notion-x/src/styles.css
+++ b/packages/react-notion-x/src/styles.css
@@ -87,6 +87,7 @@
 
   --notion-max-width: 720px;
   --notion-header-height: 45px;
+  --notion-corner-radius: 8px;
 }
 
 .dark-mode {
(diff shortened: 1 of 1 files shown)


transitive-bullshit:
these PRs are really solid. thanks, alberto 🙏

PR #688, opened 2026-03-07
feat: client-side collection load limits with 'Load more' button

albertodainotti:
## Summary

Adds client-side support for Notion's `inline_collection_first_load_limit` format property. When a collection view has a load limit configured, only the first N items are shown initially, with a "Load more" button to reveal the rest.

## Changes

- **`collection-view-gallery.tsx`**: Added load limit logic with `useState` for expand state, renders "Load more" button with chevron icon when items exceed the limit
- **`collection-view-table.tsx`**: Same load limit pattern applied to table views
- **`collection-view-list.tsx`**: Same load limit pattern applied to list views
- **`styles.css`**: Added styles for `.notion-collection-load-more` button (hover effect, spacing, cursor)

## How It Works

1. Reads `inline_collection_first_load_limit.limit` from the collection view's format
2. If a limit is set and there are more items than the limit, shows only the first N items
3. Renders a "Load more" button below the collection
4. On click, reveals all items (one-time expand, no pagination)

## Motivation

Notion allows setting a "load limit" on inline collections to keep pages compact. Without this feature, all collection items are rendered immediately regardless of the Notion-configured limit, which can result in very long pages.

### DIFF
--- a/packages/react-notion-x/src/third-party/collection-view-gallery.tsx
+++ b/packages/react-notion-x/src/third-party/collection-view-gallery.tsx
@@ -1,5 +1,5 @@
 import type * as types from 'notion-types'
-import type React from 'react'
+import React from 'react'
 import { type PageBlock } from 'notion-types'
 import { getBlockValue } from 'notion-utils'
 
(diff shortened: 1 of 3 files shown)


transitive-bullshit:
these PRs are really solid. thanks, alberto 🙏

Issue #702, opened 2026-04-02
An error occurs when calling the NotionRenderer component. installHook.js:1 TypeError: Cannot read properties of undefined (reading 'replace')


p20921:
"react-notion-x": "^6.16.0",

An error occurs when calling the NotionRenderer component.
It was working normally until April 1st (KST), but the issue started occurring this morning.

installHook.js:1 TypeError: Cannot read properties of undefined (reading 'replace')
    at K (619-fd30454….js:1:147555)
    at Fe (619-fd30454….js:1:195579)
    at ro (framework-2f612445bd…b211f15a.js:1:60227)
    at Hu (framework-2f612445bd…211f15a.js:1:112685)
    at Pi (framework-2f612445bd…b211f15a.js:1:99859)
    at xi (framework-2f612445bd…b211f15a.js:1:99787)
    at _i (framework-2f612445bd…b211f15a.js:1:99650)
    at vi (framework-2f612445bd…b211f15a.js:1:96616)
    at framework-2f612445bd…b211f15a.js:1:45995
    at t.unstable_runWithPriority (framework-2f612445bd…211f15a.js:1:130400)

IttoSanzzo:
yup, it just started happening here as well

Sangkun-svg:
i got same error 😢

hades217:
same

jooseoklee2:
I got same error

zuber-dwellfi:
Cannot read properties of undefined (reading 'replaceAll')

return (
  46 |         <div className="notion__container">
> 47 |             <NotionRenderer
     |             ^
  48 |                 recordMap={recordMap}
  49 |                 fullPage={true}
  50 |                 darkMode={false}

andremaytorena:
Same, anyone find a fix?

artdimitrov:
thanks @p20921 ! that fix worked for me https://github.com/NotionX/react-notion-x/issues/702#issue-4190688977

ButzkeGWG:
just upgrade to latest version the notion libraries

Sangkun-svg:
@ButzkeGWG thanks!!!

transitive-bullshit:
please update to latest. old versions are not guaranteed to work https://github.com/NotionX/react-notion-x/releases

PR #703, opened 2026-04-02
fix: add defensive check for undefined uuid in uuidToId

ehredev:
Notion API response structure changes can cause block IDs to be undefined in some cases, leading to `undefined.replaceAll()` errors.

#### Description
uuidToId could crash with undefined.replaceAll() when Notion API returns blocks with undefined IDs due to their recent response structure changes. Added a simple fallback (uuid || '') to prevent the error.


ehredev:
This issue is regarding #702

ButzkeGWG:
upgrade my notion libraries to the latest one solved the problem

